PR #59, opened 2015-03-18
Correlation Header ID

emmenko:
@hajoeichler please review

### DIFF
--- a/docs/CLIENT.md
+++ b/docs/CLIENT.md
@@ -25,8 +25,8 @@ SPHERE CLIENT
   * [Types of responses](#types-of-responses)
   * [Error handling](#error-handling)
     * [Error types](#error-types)
-    * [Correlation Header ID](#correlation-header-id)
   * [Statistics](#statistics)
+    * [Correlation Header ID](#correlation-header-id)
 * [Logging & debugging](DEBUGGING.md)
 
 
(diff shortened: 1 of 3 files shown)


docs/CLIENT.md
```diff
@@ -449,6 +455,11 @@ Following error types are exposed:
   * `InternalServerError`
   * `ServiceUnavailable`
 
+#### Correlation Header ID
+When a request fails with `statusCode >= 400` the reponse contains a special header `x-correlation-id`. This is specially useful for the SPHERE.IO team in order to better correlate failure requests with other possible causes.
```
svenmueller: actually the `X-Correlation-ID` header is set for any kind of request (2xx, ...)

emmenko: Ah ok. And I actually saw that headers were already returned...will adjust a couple of things then

emmenko:
@svenmueller how about now?

docs/CLIENT.md
```diff
@@ -510,3 +517,8 @@ client = new SphereClient
 }
 
 ```
+
+#### Correlation Header ID
+Each response from the HTTP API contains a special header `x-correlation-id`. This is specially useful for the SPHERE.IO team in order to better correlate failure requests with other possible causes.
```
svenmueller: it's `X-Correlation-ID`

emmenko: It's lowercase when you inspect the JSON object. Is that a problem?

svenmueller: i would remove one of the "special"/"specially" in the description. Maybe something like this:

"This unique value can be used to correlate events across different layers and also might help in case of failures."

src/spec/integration/channels.spec.coffee
```diff
@@ -29,11 +29,15 @@ describe 'Integration Channels', ->
...
-    @client = new SphereClient config: Config
+    @client = new SphereClient
+      config: Config
+      stats:
+        includeHeaders: true
 
     @client.channels.save(newChannel())
     .then (result) =>
       expect(result.statusCode).toBe 201
+      expect(result.http.response.headers['x-correlation-id']).toBeDefined()
```
svenmueller: we should use the same case for the header as we return from nginx (`X-Correlation-ID`)

hajoeichler:
:+1:

Issue #18, opened 2014-04-06
Add support for other customer endpoints


emmenko:
- [x] login
- [ ] password
- [ ] email

PR #68, opened 2015-05-26
Add AuthenticationService to log in customers

Sikian:
As the service is directly extended from BaseService, the request is performed with .save (as it's the method which will perform a json POST to the endpoint).

It could be a good idea to add an .save alias for AuthenticationService.

### DIFF
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -1,6 +1,7 @@
 _ = require 'underscore'
 Rest = require './connect/rest'
 TaskQueue = require './task-queue'
+AuthenticationService    = require './services/authentication'
 CartDiscountService      = require './services/cart-discounts'
 CartService              = require './services/carts'
 CategoryService          = require './services/categories'
@@ -25,6 +26,7 @@ TaxCategoryService       = require './services/tax-categories'
 ZoneService              = require './services/zones'
 
 ALL_SERVICES = [
+  {key: 'authentication',     name: AuthenticationService},
   {key: 'cartDiscounts',      name: CartDiscountService},
   {key: 'carts',              name: CartService},
   {key: 'categories',         name: CategoryService},
--- a/src/coffee/services/authentication.coffee
+++ b/src/coffee/services/authentication.coffee
@@ -1,6 +1,13 @@
 BaseService = require './base'
 
 # Login service as defined in http://dev.sphere.io/http-api-projects-customers.html#authenticate-customer
+#
+# Example
+#   service = client.authentication
+#   service.save
+#     email: "test@test.com"
+#     password: "1234"
+#
 class AuthenticationService extends BaseService
 
   # Internal: {String} The HTTP endpoint for `Authentication`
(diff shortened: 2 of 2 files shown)


hajoeichler:
@Sikian thx for the PR. Maybe simply add the alias like https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/services/base.coffee#L394

Could you also provide a spec, please.

src/coffee/client.coffee
```diff
@@ -1,6 +1,7 @@
 _ = require 'underscore'
 Rest = require './connect/rest'
 TaskQueue = require './task-queue'
+AuthenticationService    = require './services/authentication'
```
emmenko: I would name it `CustomerAuthenticationService`, otherwise it's a bit misleading

src/coffee/services/authentication.coffee
```diff
@@ -0,0 +1,16 @@
+BaseService = require './base'
+
+# Login service as defined in http://dev.sphere.io/http-api-projects-customers.html#authenticate-customer
+#
+# Example
+#   service = client.authentication
+#   service.save
```
emmenko: I would add an alias `login` and use only that

src/coffee/services/authentication.coffee
```diff
@@ -0,0 +1,16 @@
...
+#     email: "test@test.com"
+#     password: "1234"
+#
+class AuthenticationService extends BaseService
+
+  # Internal: {String} The HTTP endpoint for `Authentication`
+  @baseResourceEndpoint: '/login'
+
```
emmenko: Also you should mark other methods as unsupported (e.g.: http://dev.sphere.io/http-api-projects-customers.html#authenticate-customer)

`fetch`, `save`, `create`, `update`, `delete`

emmenko:
And tests please!

src/coffee/client.coffee
emmenko: I actually wonder why the API endpoint was named like this, since it's related to customer...

emmenko:
Refs https://github.com/sphereio/sphere-node-sdk/issues/18

src/coffee/services/authentication.coffee
emmenko: Sleeping it over, I think it would make more sense to have this endpoint in the `customers` service
https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/services/customers.coffee

Simply

```coffeescript
login: (body) ->
  unless body
    throw new Error 'Body payload is required for login the customer'
  @_save('/login', body)
```

Sikian: I really like this solution. I changed our code to use this idea while we get the idea you proposed merged!

hajoeichler:
@Sikian are you going to update your PR to implement @emmenko's idea from https://github.com/sphereio/sphere-node-sdk/pull/68/files#r31109950

Or should we do it?

Sikian:
I think the most sensible thing would be for your team to implement it as
you have a deeper knowledge of how the module is structured and documented.

Thanks!

Issue #75, opened 2015-07-09
Allow to process search results in chunks


hajoeichler:
see https://github.com/sphereio/sphere-node-sdk/commit/d07b0e0638d453c39a9b53111fb064cdc74f07c8#commitcomment-12091883

PR #76, opened 2015-07-09
Allow to process search results. Refs #75.

hajoeichler:
@emmenko please review

### DIFF
--- a/src/coffee/services/product-projections.coffee
+++ b/src/coffee/services/product-projections.coffee
@@ -250,7 +250,6 @@ class ProductProjectionService extends BaseService
   #   .search()
   search: ->
     @_currentEndpoint = '/product-projections/search'
-    @fetch()
 
   ###*
    * Query suggestions based on search keywords (used e.g. for auto-complete functionality)
--- a/src/coffee/services/product-projections.coffee
+++ b/src/coffee/services/product-projections.coffee
@@ -249,6 +249,23 @@ class ProductProjectionService extends BaseService
   #   .filter('variants.attributes.foo:"bar"')
   #   .search()
   search: ->
+    @asSearch()
+    @fetch()
+
+
+  # Public: Define to use search instead of query endpoint.
+  # Compared to {search} this allows for further chaining.
+  # For example to use {process} to deal with product search results in batches.
+  #
+  # Returns a chained instance of `this` class
+  #
+  # Examples
+  #
+  #   service = client.products
+  #   service
+  #   .asSearch()
+  #   .process(...)
+  asSearch: ->
     @_currentEndpoint = '/product-projections/search'
 
   ###*
--- a/src/coffee/services/product-projections.coffee
+++ b/src/coffee/services/product-projections.coffee
@@ -249,9 +249,7 @@ class ProductProjectionService extends BaseService
   #   .filter('variants.attributes.foo:"bar"')
   #   .search()
   search: ->
-    @asSearch()
-    @fetch()
-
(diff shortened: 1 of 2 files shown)


src/coffee/services/product-projections.coffee
```diff
@@ -250,7 +250,6 @@ class ProductProjectionService extends BaseService
   #   .search()
   search: ->
     @_currentEndpoint = '/product-projections/search'
-    @fetch()
```
emmenko: I don't like it, it's a breaking change. We should think of something else.

hajoeichler: It's a matter of documentation ;)

See https://github.com/sphereio/sphere-node-sdk/issues/75

emmenko: Actually it should work https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/services/base.coffee#L329
Let me have a look.

emmenko: Sorry, the endpoint is set on the fly...let me have another look then ;)

emmenko: I guess the painless solution would be to have a new method called `processSearch`. Any other solution would be a breaking change I guess. 
What do you think?

hajoeichler: I would vote to break here in order to not get to much confused.
The API is clearer afterwards.

Bump minor version + inform on release nodes + in the JSDoc.

emmenko: So something like this?

```javascript
client.productProjections.search().fetch()
client.productProjections.search().process()

// instead of
client.productProjections.search()
client.productProjections.processSearch()
```

I don't really like `search().fetch()` to be honest...feels like you are doing 2 requests.
If `search` is just used to set the endpoint, then I would rather have something like this

```javascript
client.productProjections.forSearch().fetch()
client.productProjections.forSearch().process()

// same for suggest
client.productProjections.forSuggest().fetch()
client.productProjections.forSuggest().process()
```

And then keep `search` and `suggest` for backwards compatibility.

hajoeichler: I looked into `suggest` - it does not have any paging.

emmenko: Hmm ok, well just for `search` then, so using `forSearch`.

hajoeichler: `useSearch` maybe?

emmenko: You're **using** the service **for** search. Or `asSearch`.

hajoeichler: `asSearch` FTW!

emmenko: :)

hajoeichler:
Will adapt to `asSearch` as extra method.

hajoeichler:
@emmenko please check again.

src/coffee/services/product-projections.coffee
```diff
@@ -249,9 +249,25 @@ class ProductProjectionService extends BaseService
...
+  # Returns a chained instance of `this` class
+  #
+  # Examples
+  #
+  #   service = client.products
+  #   service
+  #   .asSearch()
+  #   .process(...)
+  asSearch: ->
+    @_currentEndpoint = '/product-projections/search'
```
emmenko: You have to return `this`

emmenko:
Add a test and squash your commits then before merging, thanks

PR #85, opened 2015-08-31
Deletion of shippingMethod is possible now.

mmoelli:


mmoelli:
@emmenko please review. Is this enough?

emmenko:
@mmoelli also here https://github.com/sphereio/sphere-node-sdk/blob/master/src/spec/client/client.spec.coffee#L97

PR #88, opened 2015-10-08
Querystring

hajoeichler:
With this PR, we store each parameter from the querystring which is necessary if we use the process function for batch processing.

I used an explicit array for the plain attributes that don't need encoding.

@emmenko please review.

BTW: I don't know why only one search suggest job is failing in travis: https://travis-ci.org/sphereio/sphere-node-sdk/jobs/84246626

### DIFF
--- a/src/coffee/services/base.coffee
+++ b/src/coffee/services/base.coffee
@@ -40,6 +40,7 @@ class BaseService
     # Private: Container that holds request parameters such `id`, `query`, etc
     @_params =
       encoded: ['where', 'expand', 'sort']
+      plain: ['perPage', 'page']
       query:
         where: []
         operator: 'and'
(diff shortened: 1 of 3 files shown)


src/coffee/services/base.coffee
```diff
@@ -40,6 +40,7 @@ class BaseService
     # Private: Container that holds request parameters such `id`, `query`, etc
     @_params =
       encoded: ['where', 'expand', 'sort']
+      plain: ['perPage', 'page']
```
emmenko: What's that for?

hajoeichler: To know what parameters are also possible that are not encoded.
I use the array to loop trough when analyzing the queryString.

hajoeichler: To know what parameters are also possible that are not encoded.
I use the array to loop trough when analyzing the queryString.

PR #89, opened 2015-10-10
can configure base urls, not only hosts

yanns:
client ID and secret are not necessary when the oauth token is given

@emmenko please review

### DIFF
--- a/src/coffee/connect/oauth2.coffee
+++ b/src/coffee/connect/oauth2.coffee
@@ -25,6 +25,7 @@ class OAuth2
   # options - An {Object} to configure the service
   #   :config - {Object} It contains the credentials `project_key`, `client_id`, `client_secret`
   #   :host - {String} The host (default `auth.sphere.io`)
+  #   :protocol - {String} The protocol (default `https`)
   #   :accessTokenUrl - {String} The url path to the token endpoint (default `/oauth/token`)
   #   :user_agent - {String} The request `User-Agent` (default `sphere-node-connect`)
   #   :timeout - {Number} The request timeout (default `20000`)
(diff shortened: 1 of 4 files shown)


hajoeichler:
:+1:

yanns:
I impacted the coverage severely... :chart_with_downwards_trend:

emmenko:
@yanns I don't like this change, it's also a breaking change. Why not just allowing to configure the `protocol`? 
Ok for the `access_token` check.

yanns:
@emmenko I choose this way to be consistent with other sdks:
- the JVM SDK uses `url`: 
  http://sphereio.github.io/sphere-jvm-sdk/javadoc/v1.0.0-M18/io/sphere/sdk/client/SphereApiConfig.html#of-java.lang.String-java.lang.String-
  http://sphereio.github.io/sphere-jvm-sdk/javadoc/v1.0.0-M18/io/sphere/sdk/client/SphereAuthConfig.html#of-java.lang.String-java.lang.String-java.lang.String-java.lang.String-
- the PHP SDK uses `apiUrl ` and `oauthUrl`: http://sphereio.github.io/commercetools-php-sdk/docs/master/class-Commercetools.Core.Config.html

Considering the breaking change:
- on the input (the config object), it is not a breaking change. We sill support configuring only the hosts
- on the output, you're right: the SDK does not deliver the `host` but the `baseUrl`.

What we could do:
- change the `baseUrl` into `apiUrl ` and `oauthUrl`
- parse the `apiUrl` to extract the host and send it in the response not to break existing clients using `host`.

hajoeichler:
I like when it's consistent with all SDKs.
Please be aware this this configuration is going to be used more often now because of the new datacenter.
Also, people from the backend do run there instance locally without SSL and they should not struggle with this config anymore.

emmenko:
I'd rather use the `url` options in the new sdk.

Instead I would just add keep `host` and add `protocol` as new option. Then you have what you want.

yanns:
@emmenko why not `url` now? The code is almost done, and can be entirely backwards compatible?

emmenko:
I feel it's kind of redundant and might be confusing. At the end `baseUrl === "#{protocol}://#{host}"`. So why not simply make the protocol configurable? 

Anyway, if you want the url, just use the url.

yanns:
@emmenko I understand that having both hosts and url is quite confusing.
OK, I'll add a protocol instead.

emmenko:
:+1: thanks

yanns:
We still have problems with cloud queues. Therefore, the build should fail for tests involving the search.

yanns:
@emmenko can you make a new release with this change? I need that in impex.

emmenko:
@yanns sure, will do the release

PR #90, opened 2015-10-24
Basic support of graphql endpoint

emmenko:
/cc @OlegIlyenko

### DIFF
--- a/src/coffee/services/base.coffee
+++ b/src/coffee/services/base.coffee
@@ -3,7 +3,7 @@ _ = require 'underscore'
 _.mixin require('underscore-mixins')
 Promise = require 'bluebird'
 Utils = require '../utils'
-{HttpError, SphereHttpError} = require '../errors'
+{HttpError, GraphQLError, SphereHttpError} = require '../errors'
 
 # Private: RegExp to parse time period for last function.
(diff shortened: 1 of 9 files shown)


src/coffee/services/base.coffee
```diff
@@ -556,7 +556,14 @@ class BaseService
 
       # TODO: check other possible acceptable codes (304, ...)
       if 200 <= response.statusCode < 300
-        resolve _.extend responseJson,
+        # FIXME: find a better solution
+        if @constructor.name is 'GraphQLService' and not body.data
+          graphqlError = new GraphQLError 'GraphQL error', _.extend responseJson, body,
+            statusCode: 400
+            originalRequest: originalRequest
+          return reject graphqlError
```
emmenko: @OlegIlyenko I noticed that the endpoint return always `200` even if there are `errors`. Is that by design?

OlegIlyenko: hmm.. it's a good point. It's definitely not the part of the spec since it does not cover an HTTP interface. One can still get most of the response data even in presence of errors, so in thin case I would still  return 200. Some errors happen much earlier though, like query validation errors. I think we mat return 400 in these cases.

PR #91, opened 2015-11-08
WIP  added searchKeywords update action

sebbulon:
Hi,
I added setSearchKeywords update action to the Product Sync.
Plz review @emmenko @hajoeichler 

thanks

for some reason this test fails though- although absolutely nothing has changed in that function - so it must have been an existing issue which looks more backend related:

```
  1) Integration Products should search for suggestions
   Message:
     Expected [  ] to equal [ { text : 'Schweizer Messer' } ].
```

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -299,32 +299,33 @@ class ProductUtils extends BaseUtils
         _.each keys, (k) =>
           # we pass also the value of the correspondent key of the original object
           # in case we need to patch for long text diffs
-          console.log("got k " + k)
+          # console.log("got k " + k)
           diffed_value = obj[k]
           if _.isArray diffed_value
             value = @getDeltaValue(diffed_value, old_obj[key][k])
-            console.log("got delta value " + JSON.stringify(value))
-            updated[k] = value
+            # console.log("got delta value array " + JSON.stringify(value))
+            updated[k] = value unless _.find value, (val) ->
+              _.has(val, 'text') and val['text'] is ""   #remove empty text
           else if _.isObject(diffed_value)
             # ok this is not an array - likely the searchKeywords - removing the garbage
-            console.log "got diffed value " + JSON.stringify(diffed_value)
+            # console.log "got diffed value " + JSON.stringify(diffed_value)
             if _.has(diffed_value, '_t') and diffed_value['_t'] is 'a'
               diffed_keywords = []
               diffed_keys = _.keys diffed_value
               _.each diffed_keys, (v) ->
-                console.log "got v " + JSON.stringify(v)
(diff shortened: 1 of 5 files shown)


hajoeichler:
The test case fails because your project does not have the correct language settings.

hajoeichler:
We need to take care about the white/black listing of attribute changes.
This is necessary eg. if the CSV later contains the column or not. Otherwise we would delete the searchKeywords just because they weren't present in the CSV.

emmenko:
FYI: the new sdk has already support for search keywords.
https://github.com/sphereio/sphere-node-sdk/blob/rewrite-2.0/lib/sync/product-actions.js#L19
https://github.com/sphereio/sphere-node-sdk/blob/rewrite-2.0/test/sync/product-sync-base.spec.js#L126-L184

The mapping function did change a bit though and got simplified because of the special structure of keywords.
https://github.com/sphereio/sphere-node-sdk/blob/rewrite-2.0/lib/sync/product-actions.js#L359-L381

sebbulon:
@hajoeichler ok this would have to be done in the importer right? Not in the SDK here right? 
@emmenko thanks but is the 2.0 ready to be used? Are there any interface differences? I need the search keywords update action in the CSV Product Synch and switching to 2.0 sounds like a risk to me at this point. 

Are you happy for me to merge the change?

emmenko:
I'm not saying to switch to 2.0 (not yet), just wanted to point you to the implementation in case you needed help.

sebbulon:
@emmenko ok - it seems like your code seems to automatically merge right? The one I implemented is simply replacing based on the new object, so the person who uses this needs to take care of the merging themselves. In the typical CSV importscenario  you need to give the search keywords for all locales. I guess I should mention that in the release notes.

sebbulon:
actually, wait.. I just realised we need to merge..

sebbulon:
@emmenko @hajoeichler ok I added the delta encoding for search keywords. unfortunately it's messy due to the structure of the search keywords as I need to invalidate the update action if nothing to update, and need to filter out empty text elements .. now tests are failing, will investigate why..

emmenko:
It's tricky with the search keywords. I've managed to solve it with the patch, but make sure to have the same config that I have if you want to use it, otherwise it won't work (I'm talking also about diffpatcher config / version).

sebbulon:
@hajoeichler @emmenko it seems like the tests fail randomly :( when I increase the timeout, they go all through apart from 1) Integration Products should search for suggestions.. but nothing has changed there 

What would you suggest I should do? I desparately want to merge my new feature

PhilippSpo:
So tests are working now again. Just the code coverage that went down..

sebbulon:
:boom: ! I knew it, my code was correct, and that the world was the problem! ;) 
I guess I can now merge it right?

PR #96, opened 2015-12-03
feat(productProjectionSearch): add support for the fuzzy parameter

PhilippSpo:

PR #98, opened 2015-12-07
Feature/add category order hint

PhilippSpo:


PhilippSpo:
I accidentally create this branch from the fuzzy search branch. So the pr #96 would need to get merged first..

PhilippSpo:
@emmenko @hajoeichler 
The changes regarding fuzzy search are already merged to master. So the only file that is really important is `products.coffee` where I added an action for `setCategoryOrderHint` and also manipulated the `setCategoryOrderHint`-action after it was generated in order to be conform with the format that the action needs to have.
This is not good since we are specifically making an exception for `setCategoryOrderHint` here, but I guess we have to live with it for the moment..

emmenko:
@PhilippSpo Please do a rebase so that we just see the related changes

PhilippSpo:
@emmenko done. Should have done that in the first place sry

emmenko:
Add some tests please ;)

PhilippSpo:
What is this [line](https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/sync/category-sync.coffee#L31) for? It does not get tested and does not get called anywhere (even not from impex and the csv-sync)..

src/coffee/sync/utils/product.coffee
```diff
@@ -92,6 +92,12 @@ class ProductUtils extends BaseUtils
     actions = []
     _.each actionsBaseList(), (item) =>
       action = @_buildBaseAttributesAction(item, diff, old_obj)
+      if action && action.action == "setCategoryOrderHint"
```
hajoeichler: Please have a look at Coffeescripts shortcuts like `is`

src/coffee/sync/utils/product.coffee
```diff
@@ -92,6 +92,12 @@ class ProductUtils extends BaseUtils
...
+        if _.isUndefined action.categoryOrderHints
+          return
+        action.categoryId = Object.keys(action.categoryOrderHints)[0]
```
hajoeichler: Is there always a first element?

src/spec/integration/products-sync.spec.coffee
```diff
@@ -129,6 +151,19 @@ describe 'Integration Products Sync', ->
...
     @client.products.create(OLD_PROD)
     .then (result) =>
+      debug 'assing product to test category'
+      @client.products.byId(result.body.id).update(
+        version: result.body.version
+        actions: [
+          {
+            action: 'addToCategory'
+            category:
+              typeId: "category",
```
hajoeichler: Single quotes.

src/coffee/sync/utils/product.coffee
PhilippSpo: As long as the `_buildBaseAttributesAction` method works there will be a first element. Otherwise there would not be a diff in the `categoryOrderHints`, in which case we would not get here anyways.
@hajoeichler Should I add some safety checks anyways?

PhilippSpo:
@hajoeichler @emmenko is there something left to do here or can we merge this?

emmenko:
LGTM

hajoeichler:
LGTM, too

PR #103, opened 2016-01-28
feature(productProjection): add "old" process method

PhilippSpo:
copied it from here: https://github.com/sphereio/sphere-node-sdk/blob/v1.3.7/src/coffee/services/base.coffee#L321
adaptions:
- use createdAt for sorting, since sorting by ID is not supported by the search endpoint
- fallback to the process function of base if the endpoint is not 'search'

@emmenko @hajoeichler

### DIFF
--- a/src/coffee/services/product-projections.coffee
+++ b/src/coffee/services/product-projections.coffee
@@ -1,6 +1,7 @@
 debug = require('debug')('sphere-client')
 _ = require 'underscore'
 BaseService = require './base'
+Promise = require 'bluebird'
 
 # Public: Define a `ProductProjectionService` to interact with the HTTP [`product-projections`](http://dev.sphere.io/http-api-projects-products.html) endpoint.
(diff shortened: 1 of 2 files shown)


src/coffee/services/product-projections.coffee
```diff
@@ -319,4 +319,88 @@ class ProductProjectionService extends BaseService
...
+          queryString = @_queryString()
+
+          @_get("#{endpoint}?#{queryString}")
+          .then (payload) ->
+            fn(payload)
+            .then (result) ->
+              newTotal = payload.body.total
+              if not total or total is newTotal
+                nextPage = page + 1
+              else if total < newTotal
```
PhilippSpo: @emmenko should I add tests for handling products being added/removed during the process?

PhilippSpo:
I just talked to @OlegIlyenko and @agourlay about using the search-endpoint for exporting. As it turns out this is not a good and performant way of exporting and we should not encourage making use of the search-endpoint as a way to get loads of data out of CTP.

PR #104, opened 2016-02-01
fix(productSync): prevent duplicate actions

PhilippSpo:
@emmenko @hajoeichler

### DIFF
--- a/src/coffee/sync/utils/base.coffee
+++ b/src/coffee/sync/utils/base.coffee
@@ -47,4 +47,17 @@ class BaseUtils
         else
           throw new Error "Got unsupported number #{arr[2]} in delta value"
 
+
+  removeDuplicateActions: (actions) ->
+    _.reduce(actions, (filteredActions, action) ->
+      notUnique = _.find(filteredActions, (filteredAction) ->
+        return _.isEqual(filteredAction, action)
+      )
+      if notUnique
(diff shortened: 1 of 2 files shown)


emmenko:
@PhilippSpo what's this for?

PhilippSpo:
@emmenko I will add a test and you will see

PhilippSpo:
@emmenko So the duplicate actions were generated when using the price id for comparing two prices, since it would generate deltas like this one, when you want to replace one price with another:
```coffee
# this delta would break the price import and produce duplicate actions
delta =
  masterVariant:
    prices:
      0: [
        value:
          currencyCode: "EUR"
          centAmount: 200
      ]
      _t: "a"
      _0: [
        value:
          currencyCode: 'US'
          centAmount: 100
      , 0, 0
      ]
```
This generates the following actions:
```js
[ { action: 'removePrice',
    variantId: 1,
    price: { value: [Object] } },
  { action: 'removePrice',
    variantId: 1,
    price: { value: [Object] } },
  { action: 'addPrice', variantId: 1, price: { value: [Object] } },
  { action: 'addPrice', variantId: 1, price: { value: [Object] } } ]
```
Since we decided to not use price.id for comparing after all this is probably irrelevant an can be closed.

emmenko:
Ok

PR #107, opened 2016-02-01
fix(baseService): use limit of 20 if limit is 0

### LINKED ISSUE #106: Ensure to use a default limit value if perPage is set to 0 (process function)
Refs #105

PhilippSpo:
Since the API was changed, so that limit=0 won't return any results
closes #106 
@emmenko @mmoelli


emmenko:
:+1:

PR #111, opened 2016-02-15
feat(baseService): add secure option to filter for http authorization header

### LINKED ISSUE #109: Filter sensitive data of http headers
Use Case:
An API Error occurs during a product-import script and get's logged to the console, which gets forwarded to logentries and displayed there. Since the log contains the HTTP headers, sensitive information like the authorization key will be displayed in logentries.

Sensitive information like the authorization key should be censored automatically. This should be possible by providing a option to the sphereClient constructor.

PhilippSpo:
closes #109 
@emmenko @butenkor

### DIFF
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -134,7 +134,7 @@ ALL_SERVICES = [
 #
 # Current options are available:
 # - `includeHeaders` will include some HTTP header information in the response, wrapped in a JSON object called `http`
-# - `secure` if set to false, it will filter out all http authorization headers from the response
+# - `maskSensitiveHeaderData` if set to false, it will filter out all http authorization headers from the response
 #
 # Each response from the HTTP API contains a header `x-correlation-id`.
 # This unique value can be used to correlate events across different layers and also might help in case of failures.
@@ -145,7 +145,7 @@ ALL_SERVICES = [
 #   config: # credentials
 #   stats:
 #     includeHeaders: true
-#     secure: false
+#     maskSensitiveHeaderData: false
 # client.products.fetch()
 # .then (result) ->
 #   # result.statusCode
@@ -167,7 +167,7 @@ ALL_SERVICES = [
 #     task: {} # optional TaskQueue instance
 #     stats:
 #       includeHeaders: true
-#       secure: false
+#       maskSensitiveHeaderData: false
 #   client.products
 #   .where('name(en="Foo")')
 #   .where('id="1234567890"')
@@ -233,7 +233,7 @@ class SphereClient
       _task: @_task
       _stats: _.defaults options.stats or {},
         includeHeaders: false
-        secure: true
+        maskSensitiveHeaderData: true
 
     # TODO: currently instances are bound to the client as properties (e.g.: client.products)
     # We may think to provide a better interface for that, so for now we keep it like this
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -134,6 +134,7 @@ ALL_SERVICES = [
 #
 # Current options are available:
 # - `includeHeaders` will include some HTTP header information in the response, wrapped in a JSON object called `http`
+# - `secure` if set to false, it will filter out all http authorization headers from the response
 #
 # Each response from the HTTP API contains a header `x-correlation-id`.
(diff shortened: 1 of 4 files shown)


src/coffee/client.coffee
```diff
@@ -134,6 +134,7 @@ ALL_SERVICES = [
 #
 # Current options are available:
 # - `includeHeaders` will include some HTTP header information in the response, wrapped in a JSON object called `http`
+# - `secure` if set to false, it will filter out all http authorization headers from the response
```
butenkor: Did you mean: It will NOT filter out...?

butenkor: i would rather call it as ```filterSensitiveHeaderData```

butenkor:
I would just rename option but rest looks good.

src/coffee/client.coffee
emmenko: The correct name would be `maskSensitiveData`.

src/coffee/services/base.coffee
```diff
@@ -546,6 +550,8 @@ class BaseService
       options.config = _.omit(options.config, ['client_id', 'client_secret'])
 
     if options
```
emmenko: I think this is always `true`

src/coffee/services/base.coffee
```diff
@@ -528,6 +528,10 @@ class BaseService
   # response - {Object} An `http.IncomingMessage` object containing all kind of information about the request / response
   # body - {Object} A JSON object containing the HTTP API resource or error messages
   _wrapResponse: (resolve, reject, originalRequest, error, response, body) ->
+    if @_stats.secure is false and response
+      response.req._header = @_censorHeaderStr(response.req._header)
+      response.request.headers = @_censorHeaderObj(response.request.headers)
+      response.headers = @_censorHeaderObj(response.headers)
```
emmenko: I would not keep those functions within the class instance. Move them outside.

src/coffee/services/base.coffee
PhilippSpo: you mean options?

src/coffee/client.coffee
butenkor: :+1:

src/coffee/services/base.coffee
emmenko: Yep, you define it a few lines above

src/coffee/client.coffee
```diff
@@ -230,6 +233,7 @@ class SphereClient
       _task: @_task
       _stats: _.defaults options.stats or {},
         includeHeaders: false
+        maskSensitiveHeaderData: true
```
butenkor: I thought by default you wanted to have sensitive data in the logs?

src/coffee/client.coffee
```diff
@@ -134,6 +134,7 @@ ALL_SERVICES = [
 #
 # Current options are available:
 # - `includeHeaders` will include some HTTP header information in the response, wrapped in a JSON object called `http`
+# - `maskSensitiveHeaderData` if set to false, it will filter out all http authorization headers from the response
```
butenkor: In my opinion this sounds better:

```If set to true, it will filter out all sensitive data like authorisation from response headers. By default it is set to false.````

File: src/coffee/services/base.coffee [hunk 6]
```diff
@@ -528,6 +528,12 @@ class BaseService
   # response - {Object} An `http.IncomingMessage` object containing all kind of information about the request / response
   # body - {Object} A JSON object containing the HTTP API resource or error messages
   _wrapResponse: (resolve, reject, originalRequest, error, response, body) ->
+    if @_stats.maskSensitiveHeaderData is false and response
```
butenkor: Should be ```if @_stats.maskSensitiveHeaderData is true and response```, right?

src/coffee/services/base.coffee
```diff
@@ -545,8 +551,9 @@ class BaseService
     if options.config
       options.config = _.omit(options.config, ['client_id', 'client_secret'])
 
-    if options
-      originalRequest.options = options
+    if @_stats.maskSensitiveHeaderData is false
```
butenkor: same here

butenkor:
:+1:

PR #115, opened 2016-02-19
fix(productSync): prevent actions that would remove images from CDN unwantedly

### LINKED ISSUE #114: Prevent Images from being deleted on the CDN
### Actual behavior
When updating the label or dimensions of an image, the SDK generates a remove and an add action for this image, since there are no update actions for images in the API. Since the image is removed, it also gets removed from the CDN and would need to be reuploaded as a new image

### Expected behavior
Images are not deleted from CDN. This is currently only possible when preventing labels and dimensions from being updated, because this would result in a remove + add action (which deletes the image from the CDN).
It should be however still be possible to change the URL of a image, and then it is fine when the image gets removed from the CDN, since we use a new image/url anyway.

This will also fix an [issue](https://github.com/sphereio/sphere-node-product-csv-sync/issues/111) in the node-product-csv-sync.

PhilippSpo:
- [x] code is unit tested
- [x] ~~public code is documented~~ (no public api changes)

closes #114 
@emmenko please review

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -381,12 +381,35 @@ class ProductUtils extends BaseUtils
     _.each images, (image, key) =>
       delete image._MATCH_CRITERIA
       if REGEX_NUMBER.test key
-        unless _.isEmpty old_variant.images
-          action = @_buildRemoveImageAction old_variant, old_variant.images[key]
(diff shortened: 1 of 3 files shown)


hajoeichler:
LGTM and special thanks to the comments that explains the diff model.

emmenko:
:+1:

PR #118, opened 2016-02-29
refactor(baseService): remove standard status code of 500

### LINKED ISSUE #113: SDK returns by default statusCode = 500 if there were no response from server
If internet connection is broken or request times out SDK will return statusCode = "500" on request to CTP API. This is confusing. It should set status code "unknown" or not set it at all.

PhilippSpo:
- [x] code is unit tested

The status code 500 was also shown if there was no response from the server (timeout), which was
confusing since getting a status code looks like there was a response from the server.

closes #113 
/cc @emmenko

### DIFF
--- a/src/coffee/services/base.coffee
+++ b/src/coffee/services/base.coffee
@@ -561,7 +561,7 @@ class BaseService
       else
         errorMessage = error
       errorResp =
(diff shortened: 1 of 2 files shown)


emmenko:
:+1:

PR #119, opened 2016-02-29
fix(pagedRequest): set hasUniqueParams option to false

### LINKED ISSUE #117: Only last of multiple query parameters remain in query, in some circumstances.
I created a request using 2 ``expand()`` parameters.

``client.productProjections.staged(true).where(sku in ("some-sku")').expand('masterVariant.prices[*].channel').expand('variants[*].prices[*].channel').all().fetch()``

### Expected behavior
The generated query contains 2 ``expand`` parameters because I called ``.expand()`` twice.

### Actual behavior

The generated query contains only 1 ``expand`` parameter.

See above.

Problem does not occur if I use ``.perPage(100)`` instead of ``.all()`` in the query.

Tracing the problem, if I use ``all()``, ``perPage`` is set to 0, the if condition at https://github.com/sphereio/sphere-node-sdk/blob/c75a5ca29121a072f776d7f2e67336ad381f412b/src/coffee/services/base.coffee#L287 evaluates to true (at least in my case...) and ``@_paged()`` is called. 

Inside this method, ``@_rest.PAGED()`` is called, and there ``_.parseQuery()`` is called:
https://github.com/sphereio/sphere-node-sdk/blob/c75a5ca29121a072f776d7f2e67336ad381f412b/src/coffee/connect/rest.coffee#L184
 but without the ``false`` parameter: https://www.npmjs.com/package/underscore-mixins#_parsequery , so the 1st ``expand``parameter disappears.

PhilippSpo:
- [x] code is unit tested
- [x] ~~code is integration tested~~

so that it is possible to for example expand multiple references

fixes #117

### DIFF
--- a/src/coffee/connect/rest.coffee
+++ b/src/coffee/connect/rest.coffee
@@ -181,7 +181,7 @@ class Rest
   PAGED: (resource, resolve) ->
     splitted = resource.split('?')
     endpoint = splitted[0]
-    query = _.parseQuery splitted[1]
+    query = _.parseQuery splitted[1], false
 
     throw new Error 'Query limit doesn\'t seem to be 0. This function queries all results, are you sure you want to use this?' if query.limit and query.limit isnt '0'
 
--- a/src/spec/connect/rest.spec.coffee
+++ b/src/spec/connect/rest.spec.coffee
@@ -256,6 +256,16 @@ describe 'Rest', ->
             withTotal: false
           done()
 
+      it 'should not overwrite duplicate url parameters', (done) ->
+        spyOn(@pagedRest, 'GET').andCallThrough()
+        @pagedRest.PAGED "/products?expand=foo&expand=bar", (e, r, b) =>
+          url = @pagedRest.GET.calls[0].args[0]
+          query = url.split('?')[1]
+          query = query.split('&')
+          expect(query[0]).toEqual('expand=foo')
+          expect(query[1]).toEqual('expand=bar')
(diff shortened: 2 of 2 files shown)


PhilippSpo:
@emmenko can you hint me to the right place to add tests for this?

emmenko:
@PhilippSpo something like this https://github.com/sphereio/sphere-node-sdk/blob/master/src/spec/client/services/base.spec.coffee#L613-L616

PhilippSpo:
@emmenko That also was my initial idea, but [here](https://github.com/sphereio/sphere-node-sdk/blob/master/src/spec/client/services/base.spec.coffee#L64) the rest is mocked completely. I would have to unit test the rest module itself, not the base service, right?

emmenko:
No, you just want to check the url

emmenko:
```coffeescript
spyOn(@service, '_paged')
@service.all().expand('foo').expand('bar').expand('baz').fetch()
expect(@service._paged).toHaveBeenCalledWith "#{o.path}?limit=0&expand=foo&expand=bar&expand=baz"
```

PhilippSpo:
Nope.. the `_paged` method of the `baseService` was always called with the correct url, but the `PAGED` method of the `rest` module parses this url using `_.parseQuery()` without the hasUniqueParams set to true. Thats why we have to test the `rest` module here.

emmenko:
Ah right :+1:

PR #124, opened 2016-04-29
refactor(rest): prevent overwriting of sort when using all()

### LINKED ISSUE #122: .sort() with .all() doesn't work
If we make a query with .alI() and then .sort() by some field the results are not sorted. 
i.e 

```
return client.orders
      .where(`createdAt > "${date}"`)
      .sort('createdAt', true)
      .all()
      .fetch().then((res) => {
        return res.body.results;
      });
```

Retrieves the following:
Order: 200000567, createdAt: 2016-04-13T10:01:18.660Z
Order: 200000578, createdAt: 2016-04-13T16:36:34.916Z
Order: 200000583, createdAt: 2016-04-19T19:10:59.523Z
Order: 200000558, createdAt: 2016-04-11T13:09:32.232Z
Order: 200000563, createdAt: 2016-04-12T19:36:40.710Z

If we remove the .all()

```
return client.orders
      .where(`createdAt > "${date}"`)
      .sort('createdAt', true)
      .fetch().then((res) => {
        return res.body.results;
      });
```

Retrieves the orders sorted:
Order: 200000553, createdAt: 2016-04-10T17:39:30.393Z
Order: 200000554, createdAt: 2016-04-10T17:44:23.793Z
Order: 200000555, createdAt: 2016-04-10T17:46:33.496Z
Order: 200000556, createdAt: 2016-04-10T17:50:43.261Z
Order: 200000557, createdAt: 2016-04-10T17:53:04.936Z

I guess that internally the SDK iterates over pages in order to retrieve all results as the API doesn't provide the 'all' feature but the final results IMHO should be sorted

PhilippSpo:
- [x] code is unit tested

Now you are able to provide your own sorting when doing `all()`. If you don't provide any sorting it
will still sort by id by default.

closes #122

cc @emmenko

### DIFF
--- a/src/coffee/connect/rest.coffee
+++ b/src/coffee/connect/rest.coffee
@@ -169,8 +169,6 @@ class Rest
 
   # Public: Fetch all results of a Sphere resource query endpoint in batches of pages using a recursive function.
   #
-  # Note that traversing with pagination has been optimized. It now fetches pages sorted by `id` and iterates
-  # using a query with the `id` greater then the last one of the previous page.
   #
   # resource - {String} The API resource endpoint, with query string parameters.
   # resolve - {Function} A function fulfilled with `error, response, body` arguments. Body is an {Object} of `PagedQueryResponse`.
@@ -200,8 +198,6 @@ class Rest
       debug 'PAGED where predicate: %j', wherePredicate
 
       queryParams = _.stringifyQuery(_.extend({}, params,
-        sort: encodeURIComponent('id asc')
-        limit: limit
         withTotal: false
       , wherePredicate))
 
--- a/src/spec/connect/rest.spec.coffee
+++ b/src/spec/connect/rest.spec.coffee
@@ -3,6 +3,13 @@ _.mixin require('underscore-mixins')
 {Rest} = require '../../lib/main'
 Config = require('../../config').config
 
+idCounter = 0
+uniqueId = (prefix) ->
+  id = ++idCounter + ''
+  if prefix then prefix + id else id
+resetUniqueIdCounter = () ->
+  idCounter = 0
+
 describe 'Rest', ->
 
   it 'should initialize with default options', ->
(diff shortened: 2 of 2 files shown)


src/coffee/connect/rest.coffee
```diff
@@ -169,8 +169,6 @@ class Rest
 
   # Public: Fetch all results of a Sphere resource query endpoint in batches of pages using a recursive function.
   #
-  # Note that traversing with pagination has been optimized. It now fetches pages sorted by `id` and iterates
-  # using a query with the `id` greater then the last one of the previous page.
```
emmenko: I would keep this comment, just adjust the "sorting" part.

src/spec/connect/rest.spec.coffee
```diff
@@ -3,6 +3,13 @@ _.mixin require('underscore-mixins')
 {Rest} = require '../../lib/main'
 Config = require('../../config').config
 
+idCounter = 0
+uniqueId = (prefix) ->
+  id = ++idCounter + ''
+  if prefix then prefix + id else id
+resetUniqueIdCounter = () ->
+  idCounter = 0
```
emmenko: We should add a note here why we need to have a custom implementation. Actually we can scope it within the PAGED test only.

src/coffee/connect/rest.coffee
PhilippSpo: Well .. the default sorting is not done here, but in `fetch`. So the `PAGED` method is unaware of any sorting right now. Should we maybe write that in order for the `PAGED` request to work correctly, you should provide a sorting?

PhilippSpo: How about this:
```coffee
  # Note that in order to correctly traverse with pagination you need to provide a sorting.
  # If you use the all() method and don't provide any sorting, it will use sorting by id automatically.
  # With proper sorting PAGED fetches pages and iterates
  # using a query with the `id` greater then the last one of the previous page.
```

PR #126, opened 2016-04-29
refactor(fetch): perPage(0) which corresponds to limit=0 no longer triggers a page request.

PhilippSpo:
- [x] code is unit tested

Now the all() method no longer calls perPage(0) and thus set the limit=0. This is legacy since it
was used in the past to fetch all entries from a CTP API endpoint. Now doing limit=0 is used to get
the base stats without getting any results and all() is (still) beeing used to get all results from
an CTP API endpoint.

cc @emmenko

### DIFF
--- a/src/spec/client/services/base.spec.coffee
+++ b/src/spec/client/services/base.spec.coffee
@@ -171,7 +171,7 @@ describe 'Service', ->
 
       it 'should alias \'all\' for \'perPage(0)\'', ->
         spyOn(@service, 'perPage')
-        @service.all()
+        @service.perPage(0)
         expect(@service.perPage).toHaveBeenCalledWith 0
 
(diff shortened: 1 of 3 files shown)


src/spec/client/services/base.spec.coffee
```diff
@@ -207,7 +207,7 @@ describe 'Service', ->
       it 'should not use PAGED request when queryString is set', ->
         spyOn(@restMock, 'PAGED')
         spyOn(@restMock, 'GET')
-        @service.byQueryString('limit=10').all().fetch()
+        @service.byQueryString('limit=10').perPage(0).fetch()
```
emmenko: This should stay `.all()`

emmenko:
👍

PR #127, opened 2016-04-29
feat(auth): automatically renew an access_token before it expires

### LINKED ISSUE #44: Automatically renew access_token before it expires
See old task https://github.com/sphereio/sphere-node-connect/issues/7

emmenko:
- [x] code is unit tested

fixes #44

/cc @PhilippSpo

### DIFF
--- a/src/coffee/connect/rest.coffee
+++ b/src/coffee/connect/rest.coffee
@@ -119,7 +119,15 @@ class Rest
   # Throws an {Error} if `access_token` could not be retrieved
   _preRequest: (params, callback) ->
     _req = (retry) =>
-      unless @_options.access_token
+      isTokenExpired = false
+      if @_tokenExpiration
+        expirationTime = Date.now() + (@_tokenExpiration * 1000) -
+          # Add a gap of 2 hours before expiration time.
+          (2 * 60 * 60 * 1000)
+        isTokenExpired = Date.now() > expirationTime
+
+      if not @_options.access_token or (@_options.access_token and isTokenExpired)
+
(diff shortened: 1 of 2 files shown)


src/coffee/connect/rest.coffee
```diff
@@ -119,7 +119,15 @@ class Rest
   # Throws an {Error} if `access_token` could not be retrieved
   _preRequest: (params, callback) ->
     _req = (retry) =>
-      unless @_options.access_token
+      isTokenExpired = false
+      if @_tokenExpiration
+        expirationTime = Date.now() + (@_tokenExpiration * 1000) -
```
butenkor: Why do you use current date? You should have used here the date which has been used during last token creation, right?

emmenko: No, what you get from the API when you request an `access_token` is the `expires_in` value (2 days).
The `_tokenExpiration` is that value.

What I'm doing here is to calculate the **future** date, which consists in:

```
now + expirationPeriod - gap
```

butenkor: How can future date get ever older then now if you alwas take now + 2 days?

emmenko: Ah now I get what you mean :)

Yeah you're right, I've put it in the wrong place. Thanks for catching this! 👍

emmenko: There you go https://github.com/sphereio/sphere-node-sdk/pull/127/commits/ad1e2c6ffb069fbd820eda9d19f6f8fff57d2c26

PhilippSpo:
👍

PR #134, opened 2016-06-13
Add support for generating update actions for product state transitions

PhilippSpo:
#### Changes
- check if `byId` or `byKey` were used before calling `delete`. This previously would lead to a program error in the SDK, which is also fixed now with this PR 4d2f4b3
- handle 405 errors and show descriptive error message, since the CTP API does not provide one
- fix handling of graphql 400 errors. Previously these would not occur with status code 400. Since the now do this update was needed. abae23c
- add support for generating update actions for product transitions
- add unit and integration tests for product state transition actions

#### DOD
- [x] code is unit tested

cc/ @emmenko @hajoeichler @butenkor

### DIFF
--- a/src/spec/integration/products-sync.spec.coffee
+++ b/src/spec/integration/products-sync.spec.coffee
@@ -51,9 +51,11 @@ describe 'Integration Products Sync', ->
 
   afterEach (done) ->
     debug 'Unpublishing all products'
-    @client.products.sort('id').where('masterData(published = "true")').process (payload) =>
+    @client.products.sort('id').where('masterData(published = "true")')
+    .process (payload) =>
       Promise.all _.map payload.body.results, (product) =>
-        @client.products.byId(product.id).update(updateUnpublish(product.version))
+        @client.products.byId(product.id)
+          .update(updateUnpublish(product.version))
     .then (results) =>
       debug "Unpublished #{results.length} products"
       debug 'About to delete all products'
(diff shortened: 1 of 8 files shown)


src/spec/integration/products-sync.spec.coffee
```diff
@@ -247,3 +249,129 @@ describe 'Integration Products Sync', ->
...
+      .then(({ body: { state } }) ->
+        expect(state).toEqual({
+          typeId: 'state',
+          id: secondStateId
+        })
+        done()
+      )
+      .catch (error) ->
+        done(_.prettify(error))
+        console.log(JSON.stringify(error, null, 2))
```
emmenko: You can remove this

emmenko:
LGTM 👍

hajoeichler:
So, if the desired state is not reachable from the current state, we will get a 400 from our platform.

hajoeichler:
LGTM

PhilippSpo:
@hajoeichler should we ad a test for unreachable states?

hajoeichler:
That would be great.

PhilippSpo:
can you explain why we need this? We are just forwarding whatever error the API throws for the given request, there is no logic on the SDK side for this. The test would just approve that the API is working as expected, not that the SDK is working as expected.

butenkor:
👍

hajoeichler:
@PhilippSpo Your are right, it's nothing specific here in the client.

Issue #137, opened 2016-06-22
"TypeError: Cannot read property 'id' of undefined" on product update without state


butenkor:
### Expected behaviour
If product update does not have product state but existing product had state set than there should be no error and state on existing product should not be changed
### Actual behaviour
It throws error: TypeError: Cannot read property 'id' of undefined
### Steps to reproduce the behaviour
Create products with state or set state to existing product, provide product update without state

Also see:
https://github.com/sphereio/sphere-node-sdk/commit/e840989ad526c2559f32a2880c60880180e89884#diff-a94247bc9925c68272e1e3c1213c3556R171

butenkor:
@PhilippSpo Could please take a look?

PR #138, opened 2016-06-24
Don't generate update actions when no state is provided

PhilippSpo:
- [x] code is unit tested

Since the API throws an error when you try to set the state to null, there is no sence in letting
the user run into this. It is a common use case that the state is not known and not important to
some importing service.

#137

cc @butenkor

### DIFF
--- a/src/spec/sync/utils/product.spec.coffee
+++ b/src/spec/sync/utils/product.spec.coffee
@@ -1390,6 +1390,26 @@ describe 'ProductUtils', ->
         expect(actual).toEqual(expected)
 
       it 'should build a transitionState action for a state change', ->
+        # test with { state: null }
+        newRef = _.extend({}, @NEW_REFERENCE, { state: null })
+        delta = @utils.diff @OLD_REFERENCE, newRef
+        actual = @utils.actionsMapReferences(
+          delta, @OLD_REFERENCE, newRef
+        )
+        expected = []
+        expect(actual).toEqual(expected)
+
+        # test without state
+        delete newRef.state
+        delta = @utils.diff @OLD_REFERENCE, newRef
+        actual = @utils.actionsMapReferences(
+          delta, @OLD_REFERENCE, newRef
+        )
+        expected = []
+        expect(actual).toEqual(expected)
+
+      it 'should not build a transitionState action if no state is provided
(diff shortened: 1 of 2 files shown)


src/spec/sync/utils/product.spec.coffee
```diff
@@ -1390,6 +1390,26 @@ describe 'ProductUtils', ->
...
+        # test without state
+        delete newRef.state
+        delta = @utils.diff @OLD_REFERENCE, newRef
+        actual = @utils.actionsMapReferences(
+          delta, @OLD_REFERENCE, newRef
+        )
+        expected = []
+        expect(actual).toEqual(expected)
+
+      it 'should not build a transitionState action if no state is provided
```
butenkor: it says "it should not" but it does. See line 1417

src/spec/sync/utils/product.spec.coffee
```diff
@@ -1390,6 +1390,26 @@ describe 'ProductUtils', ->
         expect(actual).toEqual(expected)
 
       it 'should build a transitionState action for a state change', ->
```
butenkor: Probably you need to swap test descriptions?

src/spec/sync/utils/product.spec.coffee
```diff
@@ -1390,6 +1390,26 @@ describe 'ProductUtils', ->
...
+        # test with { state: null }
+        newRef = _.extend({}, @NEW_REFERENCE, { state: null })
```
butenkor: If i understand it right ```newRef``` will have this structure for product ```{...{"state": null}}``` but it should be ```{...}```. If state is not known than state attribute is not available in new product at all. It should support also state with value null but in most cases state will not be set at all

PhilippSpo: There actually is a test for this case [a few lines further](https://github.com/sphereio/sphere-node-sdk/pull/138/files#diff-2a9c8ef1ff08242733bd0f96f869ada8R1403) 😉

src/spec/sync/utils/product.spec.coffee
PhilippSpo: Good catch!

butenkor:
LGTM

PR #139, opened 2016-06-27
Correct the objectHash when the value is null

adnasa:
the `jsondiffpatch` executes `objectHash` and may pass in `null` as the `obj` argument value
This PR takes `null` into account and returns the objects hash-index if that were the case.

- [x] code is unit tested

---

cc @ct-dali @PhilippSpo


lib/sync/utils/diffpatcher.js
```diff
@@ -2,7 +2,9 @@ import jsondiffpatch from 'jsondiffpatch'
 
 const diffpatcher = jsondiffpatch.create({
   objectHash (obj, index) {
-    return obj.id || obj.name || obj.url || `$$index:${index}`
+    const objIndex = `$$index:${index}`
+    if (obj === null) return objIndex
```
ct-dali: no... what if `obj` is not an object? the following code will crash.

emmenko:
This is not acceptable without a test case that reproduces the problem.

test/utils/diffpatcher.spec.js
```diff
@@ -0,0 +1,94 @@
...
+          .map((r, i) => objectHash(r, i))
+
+        t.deepEqual(objectHashIndexValues, [
+          'random1',
+          '$$index:1',
+          '$$index:2',
+          '$$index:3',
+          '$$index:4',
+        ], 'expect index value OR hash of the object in the list')
+        t.end()
```
adnasa: here is the test coverage

cc @emmenko

emmenko: Sorry, this is not exactly what I meant. A test with product data that reproduces the error, using buildActions main function. See sync tests, or ask @PhilippSpo

PhilippSpo:
I don't understand why the coverage went down...

emmenko:
@adnasa @PhilippSpo what's the status here?

PhilippSpo:
Coverage of untouched files went down

emmenko:
Don't worry about the coverage

PhilippSpo:
The only thing we could add is an integration test then.

adnasa:
I'll try to find the time today to figure _why_ test coverage went down.
Other than that, I think we are good :)
cc @PhilippSpo

adnasa:
I'd like to have this a go
cc @PhilippSpo @emmenko 

I think one of the main reasons that coverage went down is because more code is added and in the case of http and task-queue utils, some utility were not covered.

Trying to cover them would be out of scope of this PR

emmenko:
🚀

adnasa:
🍹

cc @PhilippSpo @emmenko

PR #141, opened 2016-07-11
fix(product-sync): migrate deprecated set sku action

PhilippSpo:
- [x] code is unit tested

cc @emmenko


emmenko:
LGTM 👍

PR #144, opened 2016-07-13
fix(product-sync): allow unsetting of sku if one was set

PhilippSpo:
#### Summary

This is a bug that was noticed in the MC. At the moment when you unset the SKU of an existing variant (that has an SKU set), there is no update action generated.
This is now changed, so that the only case where no update action is generated is when the old and the new draft have no SKU set (or an empty SKU)

#### DoD

- [x] code is unit tested

#### Reviewers

- [x] @emmenko has approved the PR
- [x] @adnasa has approved this PR

### DIFF
--- a/lib/sync/product-actions.js
+++ b/lib/sync/product-actions.js
@@ -195,7 +195,7 @@ export function actionsMapPrices (diff, oldObj, newObj) {
 function _buildSkuActions (variantDiff, oldVariant) {
   if (variantDiff.hasOwnProperty('sku')) {
     const newValue = diffpatcher.getDeltaValue(variantDiff.sku)
-    if (!newValue) return null
+    if (!newValue && !oldVariant.sku) return null
 
     return {
       action: 'setSku',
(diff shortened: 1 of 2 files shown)


adnasa:
The code coverage is def [not an issue](https://coveralls.io/builds/6993604/source?filename=lib%2Findex.js) in this case, @PhilippSpo 😉

adnasa:
LGTM, @PhilippSpo 
I even went as far as to test `undefined` on the original (before) and change it to `''` (now) just to check how `jsondiffpatch` behaves


__example__:
```js
const before = {
  id: '123',
  masterVariant: {
    id: 1, attributes: [{ name: 'foo', value: 'bar' }], // undefined field (sku in this case)
  },
}

const now = {
  id: '123',
  masterVariant: {
    id: 1, sku: '', attributes: [{ name: 'foo', value: 'bar' }], // a new field is introduced, with '' (not undefined)
  },
}
```

Testing that locally yielded an expected behaviour from the `jsondiffpatch`.

---

This is good for a merged to me 👍

PR #149, opened 2016-08-17
chore(builds): introduce UMD builds

emmenko:
- [x] code is unit tested
- [ ] ~~public code is documented~~

Extra:
- move source files into src folder


emmenko:
Running `npm run build` will generate 2 UMD files that you can use directly in the browser
- `dist/sphere-node-sdk.js`
- `dist/sphere-node-sdk.min.js`

PhilippSpo:
Should we add some documentation that shows how you can utilize this?

emmenko:
In the README? Yeah I guess so. Similar to this? https://github.com/reactjs/redux#installation

PhilippSpo:
Yes, we can basically use the paragraph about using the UMD build.

nkuehn:
cool! 

Did you consider letting it generate the source-map file, too for the minified version? (`devtool: 'source-map'`  config). 

I'm not sure about the documentation.  To me the only relevant use case are in-Browser apps but currently the SDK can only work with an existing access token in-Browser (would need to switch to header-based HTTP Basic auth at the OAuth API to work with clientId and clientSecret - username:password in the URL is not allowed by the fetch specification and XHR).

emmenko:
> Did you consider letting it generate the source-map file, too for the minified version?

We can do that, no problem.

> To me the only relevant use case are in-Browser apps but currently the SDK can only work with an existing access token in-Browser (would need to switch to header-based HTTP Basic auth at the OAuth API to work with clientId and clientSecret - username:password in the URL is not allowed by the fetch specification and XHR).

Yeah, this is actually a problem with our OAuth WS, which doesn't support the implicit client crendentials, used for browser applications.
I would **not recommend** to put the `clientId` + `clientSecret` in the browser though, unless the oauth client has really limited scopes. Still, the SDK at the moment only supports full scopes.

Idea: why don't you use the MC API instead? You still have full scopes but you don't need any oauth clients, you just need to get an "MC" auth token using the user login credentials.

nkuehn:
@emmenko  the permission model for how to access the API from the browser is there already: http://dev.commercetools.com/http-api-authorization.html#password-flow  and used in the iOS app.  
It requires a clientId+clientSecret, too (in addition to user name and password) and needs to send them in the Basic Auth base64 encoded header.  

(For my debug app it's different, but that's not the key use case imho - OK to have the state there that we had in IMPEX before we integrated the platform login.  Here an implicit flow could makes sense, but I see that as a low-prio topic in comparison to shop user login)

PR #152, opened 2016-08-23
[2.0] feat(auth): automatically renew an access_token before it expires

Arxisos:
This adds automatic renewal of the access_token to the `rewrite-2.0` branch.

When you merge `2.0-beta-middleware-api` you'll need to tweak the inner implementation, but you should be able to use more or less the same test.
- [x] code is unit tested

### DIFF
--- a/src/utils/task-queue.js
+++ b/src/utils/task-queue.js
@@ -30,6 +30,18 @@ export default options => {
   const http = httpFn(options)
 
   /**
+   * Returns true if the stored access token is valid.
+   * Otherwise a new access token should be requested.
+   */
+  function hasValidAccessToken () {
+    if (!auth.accessToken) return false
+
+    const { accessTokenExpirationTime: expirationTime } = auth
+    if (expirationTime && Date.now() > expirationTime) return false
+    return true
+  }
+
+  /**
    * Execute the given task and call the `cb` when done.
    *
    * @param  {Object}   task - It contains `fn`, `resolve`, `reject` which are
@@ -53,7 +65,7 @@ export default options => {
   const queue = async.queue((task, cb) => {
     auth.shouldRetrieveToken(shouldRetrieve => {
       if (shouldRetrieve)
-        if (auth.accessToken) execTask(task, cb)
+        if (hasValidAccessToken()) execTask(task, cb)
         else {
           queue.pause()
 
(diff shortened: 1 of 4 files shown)


src/utils/task-queue.js
```diff
@@ -53,14 +63,24 @@ export default options => {
...
+        if (hasValidAccessToken()) execTask(task, cb)
         else {
           queue.pause()
 
-          authUtils.getAccessToken(options)
+          authUtils.getAccessToken(Object.assign({}, options, {
+            auth: Object.assign({}, options.auth, {
+              accessToken: undefined,
+              accessTokenExpirationTime: undefined,
+            }),
```
emmenko: What do you need this for?

test/utils/task-queue.spec.js
```diff
@@ -49,4 +49,87 @@ test('Utils::taskQueue', t => {
...
+        t.deepEqual(res.body, { foo: 'bar' })
+
+        return taskQueue.addTask(task)
+          .then(res => {
+            t.deepEqual(res.body, { foo: 'bar' })
+            t.end()
+          })
+      })
+      .catch(t.end)
+  })
```
emmenko: Sorry, I'm a bit confused about this test. How exactly do you check that the token is being renewed?

emmenko:
@Arxisos Thanks for the PR!
Yeah, I guess things will be a bit different in the `beta` release, hopefully I'll get to continue working on that in the next weeks.

src/utils/task-queue.js
Arxisos: If I don't set `accessToken` to `undefined` here `getAccessToken` will sent the expired token in the `Authorization` header to server which results in an error, because the server thinks I want to retrieve a new token by sending the client credentials in the `Authorization` header instead of sending them in the url (http://dev.commercetools.com/http-api-authorization.html#client-credentials-flow).

test/utils/task-queue.spec.js
Arxisos: Basically the mock server (`httpMock`) expects a new access token for every subsequent non-oauth request (so it has to track the request count with `apiCalls` which is used as part of the token). The returned token expires in 1h59m59s which results in a new token being fetched for every subsequent request (the threshold for token expiration is 2h).

emmenko:
Not relevant anymore, see new implementation [here](https://github.com/commercetools/nodejs)

PR #157, opened 2016-09-08
feat(subscriptions): add subscription endpoint

lojzatran:
- [ ] code is unit tested

@emmenko @PhilippSpo @butenkor Can you review? This feature is still in testing mode. There's no docs about it. So for now, I just added a class without any comments, so it's possible to test it. Once this feature is released, it can be improved.

### DIFF
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -22,6 +22,7 @@ ProjectService           = require './services/project'
 ReviewService            = require './services/reviews'
 ShippingMethodService    = require './services/shipping-methods'
 StateService             = require './services/states'
+SubscriptionService      = require './services/subscriptions'
 TaxCategoryService       = require './services/tax-categories'
 TypeService              = require './services/types'
 ZoneService              = require './services/zones'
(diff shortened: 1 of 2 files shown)


emmenko:
As discussed with @butenkor using this endpoint doesn't necessarily require the SDK, as it's meant to be used like once or twice. Given also that it's not documented and it's still in early development I would not include it in the SDK, yet.
If you want to use it now, simply do it with curl.

butenkor:
@emmenko @PhilippSpo Can we use it as beta feature? I assume nobody will use it if it is not documented yet.

mmoelli:
@butenkor This feature was meant for internal testing only ATM. Please consult with @cneijenhuis @gogregor regarding release procedure.

butenkor:
@mmoelli How harmful is that if it is not documented and would be marked as BETA in code?

butenkor:
@lojzatran will add comment ```"FOR INTERNAL USAGE ONLY AND MAY BE REMOVED ANYTIME!" ```

PR #159, opened 2016-09-16
feat(subscriptions): add subscription endpoint

lojzatran:
- [ ] code is unit tested

See discussions about this topic here: https://github.com/sphereio/sphere-node-sdk/pull/157

### DIFF
--- a/src/coffee/services/subscriptions.coffee
+++ b/src/coffee/services/subscriptions.coffee
@@ -1,5 +1,9 @@
 BaseService = require './base'
 
+# IMPORTANT:
+# THIS FEATURE IS FOR INTERNAL USE ONLY.
+# MIGHT BE REMOVED IN THE FUTURE.
+# PLEASE DON'T USE THIS ENDPOINT.
 class Subscriptions extends BaseService
 
   # Internal: {String} The HTTP endpoint for `Subscriptions`
--- a/src/coffee/services/subscriptions.coffee
+++ b/src/coffee/services/subscriptions.coffee
@@ -1,0 +1,8 @@
+BaseService = require './base'
+
+class Subscriptions extends BaseService
+
+  # Internal: {String} The HTTP endpoint for `Subscriptions`
+  @baseResourceEndpoint: '/subscriptions'
+
+module.exports = Subscriptions
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -22,6 +22,7 @@ ProjectService           = require './services/project'
 ReviewService            = require './services/reviews'
 ShippingMethodService    = require './services/shipping-methods'
 StateService             = require './services/states'
+SubscriptionService      = require './services/subscriptions'
 TaxCategoryService       = require './services/tax-categories'
 TypeService              = require './services/types'
 ZoneService              = require './services/zones'
@@ -48,6 +49,7 @@ ALL_SERVICES = [
   {key: 'reviews',            name: ReviewService},
   {key: 'shippingMethods',    name: ShippingMethodService},
   {key: 'states',             name: StateService},
+  {key: 'subscriptions',      name: SubscriptionService},
   {key: 'taxCategories',      name: TaxCategoryService},
   {key: 'types',              name: TypeService}
   {key: 'zones',              name: ZoneService}


butenkor:
@hajoeichler @PhilippSpo @mmoelli @emmenko How about merging it?

mmoelli:
Still against it. Features get not released like this. The decision should be made by the responsible backend devs which can decide whether we already achieved the sufficient maturity level. If so, we will put into beta, but not before. Any prior release would lead to a malformed expectation management as this feature is very likely to be changed comparing to the current state.

butenkor:
@mmoelli How this feature should be tested in automated way if you have not client for it?

butenkor:
@cneijenhuis @gogregor WDYT?

gogregor:
If you really need to expose it, then do so only for subscriptions with the ironmq destination.

lojzatran:
@gogregor what do you mean by that?

gogregor:
It should not yet be visible to the public how you can create a sqs subscription.

src/coffee/services/subscriptions.coffee
```diff
@@ -0,0 +1,12 @@
+BaseService = require './base'
+
+# IMPORTANT:
+# THIS FEATURE IS FOR INTERNAL USE ONLY.
+# MIGHT BE REMOVED IN THE FUTURE.
+# PLEASE DON'T USE THIS ENDPOINT.
```
emmenko: This doesn't make much sense. If it says "don't use it", why do we need to have it? :)

Nevertheless, having this available when it's not officially documented by the API it's not really nice. I see 3 options here:
- you don't use the SDK and make the requests by yourself, the implementation is pretty simple
- you use the new SDK (alpha version), which allows you to register services by yourself (which is exactly this use case). The new version is still not fully-featured yet though, so you have to check if it's enough what there is available
- add this service here (less preferred option)

In any case, tests are missing for this PR.

butenkor: It clearly states "for internal use ONLY" so why it does not make sense?
What is the difference between this and BETA features (they can have breaking changes also but are supported by SDKs)? We have implemented already first option and i do not like it as it is definitely more work than this tiny PR and does not feel right as we are using this feature in production already.

We did not provide tests on purpose - so that nobody is tempted to try it out on your own already ;)

butenkor:
@lojzatran please provide tests as it is going officially as beta soon so we hopefully can merge it.

stmeissner:
The subscription feature had been [beta-released](https://dev.commercetools.com/release-notes.html#-release-notes---commercetools-platform---version-release-08112016) a couple months ago, what is the status here?

hisabimbola:
Closing this in favour of the new SDK. It's supported by the new SDK

lojzatran:
@hisabimbola can you reopen this PR? I'm working on the tests and will finish this soon. Thanks!

hisabimbola:
We are not implementing new features for the old sdk. Please migrate to the new one [here](https://commercetools.github.io/nodejs/)

lojzatran:
But it looks like the new SDK is still not finished yet. E.g. the documentation is still missing the CLI part (which is probably interesting for me because I'm working on CLI) https://commercetools.github.io/nodejs/cli/. And the old SDK is not deprecated yet and in the home page it still says nothing about migration to SDK, so we should still support, right 😛  ?

This feature would help many projects here in our team as we're currently using our own code for the subscription endpoint. And it might take a lot of time to migrate all the projects to new SDK.

butenkor:
@hisabimbola 
- This PR is not new. Only tests were missing as at the moment of its creation API did not provide official documentation yet.
- It does not seem user friendly to just stop providing new API features without announcing clear deprecation process (especially if feature is not even provided by owner). Deprecation process could look like that: https://developer.mozilla.org/en-US/Add-ons/SDK/Guides/SDK_API_Lifecycle#Deprecation_Process
- In my opinion it is not fair deal to ask migration from users if your own open source tools based on node-sdk are not migrated yet
- New SDK does not provide all the features of current SDK like: [maskSensitiveHeaderData](https://github.com/sphereio/sphere-node-sdk/blob/971e04e4d49d68a810314af11464902e186dfa7d/src/coffee/client.coffee#L143), [request repeater](https://github.com/sphereio/sphere-node-sdk/blob/971e04e4d49d68a810314af11464902e186dfa7d/src/coffee/client.coffee#L80), [custom timeout](https://github.com/sphereio/sphere-node-sdk/blob/971e04e4d49d68a810314af11464902e186dfa7d/src/coffee/connect/rest.coffee#L19). Probably there is more.
- As other opens source libraries based on old SDK or not migrated yet it will surely pollute our code with two different ways how we provide data for CTP [client creation]( https://github.com/sphereio/sphere-product-import/blob/master/lib/product-import.coffee#L26)

emmenko:
Ok, let's just go ahead with this

PR #162, opened 2016-09-28
fix(product-import): allow import of masterVariant

### LINKED ISSUE #161: Product sync should allow syncing products where new product does not have masterVariant defined
### Expected behavior
It should create delete action for masterVariant
### Actual behavior
It throws an exception as it validates existence of sku or id on non existing masterVariant
https://github.com/sphereio/sphere-node-sdk/blob/794c68a1eeb37bf7a33c1377d2ea8ea8081ea5dc/src/coffee/sync/utils/product.coffee#L34
Create a product with masterVariant and empty variants array. Remove from new product masterVariant and fill variants array with another variant and sku of master.

hisabimbola:
- [x] code is unit tested

fixes #161

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -68,7 +68,9 @@ class ProductUtils extends BaseUtils
 
     patch = (obj, arrayIndexFieldName) ->
       debug 'patching product: %j', obj
-      _.each allVariants(obj), (variant, index) ->
+      _allVariants = allVariants(obj)
+      _.each _allVariants, (variant, index) ->
+        return variant unless variant?
         patchPrices variant
         patchEnums variant
         patchSetLText variant
@@ -79,7 +81,6 @@ class ProductUtils extends BaseUtils
 
     patch old_obj, '_EXISTING_ARRAY_INDEX'
     patch new_obj, '_NEW_ARRAY_INDEX'
-
     super old_obj, new_obj
 
   # Private: map base product actions
@@ -624,6 +625,6 @@ actionsBaseList = ->
 
 allVariants = (product) ->
   {masterVariant, variants} = _.defaults product,
-    masterVariant: {}
+    masterVariant: undefined
     variants: []
   [masterVariant].concat variants
--- a/src/spec/sync/product-sync.spec.coffee
+++ b/src/spec/sync/product-sync.spec.coffee
@@ -190,3 +190,33 @@ describe 'ProductSync', ->
         ]
         version: oldProduct.version
       expect(update).toEqual expected_update
+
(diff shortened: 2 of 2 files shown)


src/coffee/sync/utils/product.coffee
```diff
@@ -624,6 +625,6 @@ actionsBaseList = ->
 
 allVariants = (product) ->
   {masterVariant, variants} = _.defaults product,
-    masterVariant: {}
+    masterVariant: undefined
     variants: []
   [masterVariant].concat variants
```
butenkor: isn't here better place to check if masterVariant is defined or not?

```
 allVariants = (product) ->
    variants = []
    if (product?.masterVariant)
      variants = variants.concat(product.masterVariant)
    if (product?.variants)
     variants = variants.concat(product.variants)
    variants
```

hisabimbola: Hey @butenkor 

yeah I tried that but with the way the code is setup, a masterVariant object is always expected to be in the product object and that it should be the first item in the variants array. So I set it to undefined and do not do any action on it [here](https://github.com/sphereio/sphere-node-sdk/pull/162/files#diff-a94247bc9925c68272e1e3c1213c3556R73)

butenkor: > masterVariant object is always expected to be in the product object and that it should be the first item in the variants array

You mean test are setup this way but not code, right?

hisabimbola: > You mean test are setup this way but not code, right?
The code.

butenkor: Do you have an example please?

hisabimbola:
```coffee
patch = (obj, arrayIndexFieldName) ->
      debug 'patching product: %j', obj
      _allVariants = allVariants(obj)
      _.each _allVariants, (variant, index) ->
        return variant unless variant?
        patchPrices variant
        patchEnums variant
        patchSetLText variant
        patchImages variant
        patchVariantId variant, index
        if index > 0 # this expects the first item in the product to be the masterVariant, that's why arrayIndexFieldName is not assigned. So if the master variant is not present, the first item in the variants list is not assigned this value and that causes an error somewhere else.
          variant[arrayIndexFieldName] = "#{index - 1}" # for **variants** we store the actual index in the array

    patch old_obj, '_EXISTING_ARRAY_INDEX'
    patch new_obj, '_NEW_ARRAY_INDEX'
```

@emmenko I saw that you wrote this. I am on track with this?

butenkor: I see, thanks. LGTM

PhilippSpo:
LGTM

PR #165, opened 2016-10-06
[MC-1710] Customer addresses update actions

JoshBones:
This PR adds generation of update actions for customer addresses.

Additionally, it includes a util to help generate update actions for any nested objects such as product variants, customer addresses etc.

Finally, product variant add / remove action building has been migrated to use the new utility.

- [x] code is unit tested


src/sync/utils/create-build-array-actions.js
```diff
@@ -0,0 +1,107 @@
...
+    const addActions = []
+    const removeActions = []
+    const changeActions = []
+
+    if (diff[key]) {
+      const nestedObjects = diff[key]
+      Object.keys(nestedObjects).forEach((index) => {
+        if (config[ADD_ACTIONS] && isCreateAction(nestedObjects, index))
+          addActions.push(
+            config[ADD_ACTIONS](oldObj[key], newObj[key], index)
```
emmenko: @PhilippSpo suggested to actually pass the "oldValue" and "newValue" directly, that way we don't need to pass the index anymore. This covers all current cases already.

We need to differentiate between the different cases though:

```js
// When adding a new element you don't need the oldObj
if (config[ADD_ACTIONS] && isCreateAction(nestedObjects, index))
  addActions.push(newObj[key][index])

// When removing an existing element you don't need the newObj
if (config[REMOVE_ACTIONS] && isRemoveAction(nestedObjects, index)) {
  const realIndex = index.replace('_', '')
  removeActions.push(oldObj[key][realIndex])
}

// When changing an existing element you need both
if (config[CHANGE_ACTIONS] && isChangeAction(nestedObjects, index)) {
  const oldValue = oldObj[key][index]
  const newValue = newObj[key][index]
  changeActions.push(oldValue, newValue)
}
```

src/sync/utils/create-build-array-actions.js
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Tests a delta to see if it represents a create action.
+ * eg. delta:
+ * {
+ *   0: [ { foo: 'bar' } ]
+ * }
+ *
+ * @param  {object}  obj The delta generated by the diffpatcher
+ * @param  {string}  key key of generated delta to examine
+ * @return {Boolean}     Returns true if delta reprents a create action,
```
emmenko: `represents`

src/sync/utils/create-build-array-actions.js
```diff
@@ -0,0 +1,107 @@
...
+ * eg. delta:
+ *
+ * {
+ *   0: {
+ *     foo: ['bar', 'baz']
+ *   }
+ * }
+ * @param  {object}  obj The delta generated by the diffpatcher
+ * @param  {string}  key key of generated delta to examine
+ * @return {Boolean}     Returns true if delta reprents a change action,
```
emmenko: `represents`

src/sync/utils/create-build-array-actions.js
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Tests a delta to see if it represents a remove action.
+ * eg. delta:
+ *
+ * {
+ *   _0: [ 'foo', 0, 0 ]
+ * }
+ * @param  {object}  obj The delta generated by the diffpatcher
+ * @param  {string}  key key of generated delta to examine
+ * @return {Boolean}     Returns true if delta reprents a remove action,
```
emmenko: `represents`

src/sync/utils/create-build-array-actions.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param  {string} key    key of the attribute containing the array of
+ *   nested objects
+ * @param  {object} config configuration object that can contain the keys
+ *   [ADD_ACTIONS, REMOVE_ACTIONS, CHANGE_ACTIONS], each of
+ *   which is a function. The function should accept the old + new arrays and
+ *   return an action object.
+ * @return {Array}        The generated array of actions
+ */
+export default function createBuildArrayActions (key, config) {
+  return function buildNestedObjectActions (diff, oldObj, newObj) {
```
emmenko: This function might be renamed as well

src/sync/utils/create-build-array-actions.js
```diff
@@ -0,0 +1,107 @@
...
+    const addActions = []
+    const removeActions = []
+    const changeActions = []
+
+    if (diff[key]) {
+      const nestedObjects = diff[key]
```
emmenko: Same for this variable (naming)

src/sync/utils/create-build-array-actions.js
JoshBones: done

src/sync/utils/create-build-array-actions.js
JoshBones: done

src/sync/utils/create-build-array-actions.js
JoshBones: done

src/sync/utils/create-build-array-actions.js
PhilippSpo: I guess you can argue both ways, but I think it is unnecessary cognitive load for the user to think about "Do I need the old obj here? Why would I even get it as a param then?". The implementation will be different in all three cases, so I don't see a problem with having three different signatures... I don't really have a strong opinion on this, though.

emmenko: It is simpler to pass the same arguments to those functions, on the other hand it is simpler from who's using the function to get only the useful values.

emmenko: I would refactor this, the purpose of those common functions is to simplify as much as possible the usage of them.

JoshBones:
@emmenko @PhilippSpo all request changes have been made now

PR #168, opened 2016-10-14
fix(product-actions): use correct uniq function to ensure unique sfa attributes

emmenko:
#### Summary
Fix a problem for product update actions where duplicate SFA attributes were generated.

#### Description
If multiple variants have the same SFA attribute, and this somehow changed (added, removed, changed), multiple equal actions were generated.

The problem was that the wrong lodash function was used, which allows to filter out from an array non-unique items.

#### TODO
- Tests
    - [x] Unit


emmenko:
@PhilippSpo here we go 🙈😅

emmenko:
Released in `2.0.0-alpha.44`. 

/cc @PhilippSpo

PR #179, opened 2016-11-08
Product sync custom price

Siilwyn:
- [x] code is unit tested

### DIFF
--- a/src/spec/sync/product-sync.spec.coffee
+++ b/src/spec/sync/product-sync.spec.coffee
@@ -209,3 +209,102 @@ describe 'ProductSync', ->
         ]
         version: oldProduct.version
       expect(update).toEqual expected_update
+
+    describe 'actionsMapPriceCustom', ->
+      ieNew =
+        sku: 'abc'
+        custom: {
+          type: {
+            typeId: 'type',
+            id: '123'
+          },
+          priceId: '456',
+          staged: false
+          fields: {
+            nac: 'ho'
+          }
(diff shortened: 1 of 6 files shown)


src/spec/sync/product-sync.spec.coffee
```diff
@@ -209,3 +209,102 @@ describe 'ProductSync', ->
         ]
         version: oldProduct.version
       expect(update).toEqual expected_update
+
+    describe 'actionsMapPriceCustom', ->
+      ieNew =
```
PhilippSpo: Something seems wrong here.
- Why are we passing variants and not products?
- The structure of the variant is also wrong. The custom price object is should be in the `price` and not directly in the variant.
I think this is how it should look:
```coffee
newProduct =
  id: '123',
  version: 1,
  variants: [{
    sku: 'abc',
    price: {
      id: '',
      value: {
        currenceyCode: 'EUR',
        centAmount: 4200
      },
      custom: {
        type: {
          typeId: 'type',
          id: '123'
        },
        fields: {
          nac: 'ho'
        }
      }
    }
  }]
```
The same applies for the next couple of tests.

Siilwyn:
Closed because handling custom types works be removing the whole price object and adding it which is the current implementation. 

To implement specific updating with 'setProductPriceCustomType' we need the price ID, which is not available in the scope of the mapping functions. To get this to work it would require changes to the current way our mappings are structured.

PR #181, opened 2016-11-10
Test price custom type and field handling

Siilwyn:
- [x] code is unit tested


PhilippSpo:
Wait. So everything was basically already working? 🤔

Siilwyn:
@PhilippSpo yes it is because it removes the whole price object so it also 'accidentally' includes the nested custom object.

PR #185, opened 2016-11-22
Rewrite 2.0 auth via HTTP headers

nkuehn:
This change enables the SDK to be used with clientID and client secret in a browser environment (browser's security rules don't allow `user:pass@host` notation in cross-origin requests). 

It's not working in production yet since the CT auth API does not yet send the necessary CORS headers, but serves as an enabler to motivate fully enabling the API for browser based applications. 

- [ ] code is unit tested

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage
 dist
 lib
 version.js
+.idea
--- a/src/utils/auth.js
+++ b/src/utils/auth.js
@@ -15,12 +15,23 @@ export function buildRequest (options) {
     credentials: { projectKey, clientId, clientSecret },
   } = options
 
-  const authHost = `https://${clientId}:${clientSecret}@${host}`
+  const authHost = `https://${host}`
   const endpoint = `${authHost}/oauth/token`
   const body = 'grant_type=client_credentials' +
     `&scope=manage_project:${projectKey}`
 
-  return { endpoint, body }
+  // Notes on the encoding:
+  // Since result is smaller than 76 chars the "MIME" restrictions for base64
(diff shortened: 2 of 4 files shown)


nkuehn:
Sorry, I can't make sense of the coverage feedback.  
E.g. coveralls gives this file as "coverage changed" https://coveralls.io/builds/8945943/source?filename=src%2Futils%2Ftask-queue.js   But I can't see what the code marked red has to do with my 
changes. 

Can someone help?

emmenko:
Thanks, I'll hold off on that though as those changes will be done in the auth middleware. Do you need those changes now?

nkuehn:
Since the auth API does not do it yet it's not technically urgent.  But I want to push the In-Browser case forward and without SDK we don't even find out what's necessary.  

E.g. If I hadn't done the change I wouldn't have learnt that Safari has even more special requirements (in this case: does show a big red warning if you modify the user-agent header.   

It's up to you finally, what's your timeline for "auth middleware"?   If its "sometime 2017" I would say we merge this to push the platform side of things forward. It's mostly a learning curve at the moment that goes back and forth between SDK and API.

emmenko:
I plan to push the beta version end of this year, so hopefully soon ;)

nkuehn:
OK, it's a small side job for me only.  Let's talk when you get to the implementation of "your way to do it".   The base64 encoding part is a bit messy but worth it if the resulting code really runs across both browsers and Node versions.

emmenko:
Not relevant anymore, see new implementation [here](https://github.com/commercetools/nodejs)

PR #186, opened 2016-11-23
Fix updating existing custom fields

Siilwyn:
- [x] code is unit tested

When a custom field already exists use the new object value. This solves the [breaking build](https://travis-ci.org/sphereio/sphere-stock-import/builds/178220591) in stock import. I introduced this bug when just using `diff` in 79e3475aae2569056ebe688938d46d38e8b1a8fc. 🤕


Siilwyn:
@hisabimbola yes 🏴

PR #190, opened 2016-12-05
Add default task repeater

### LINKED ISSUE #187: Client should retry on network issues or API errors by simple configuration
### Expected behavior
On network issues or 500 errors from API it should retry request
### Actual behavior
It rejects with error and caller has to implement retry logic

Possible implementation for default retry: https://gist.github.com/butenkor/93098cdd1c2185ad7102b10d9b30cd7e

LEQADA:
Fixes #187

- [x] code is unit tested

### DIFF
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -1,5 +1,6 @@
 _                        = require 'underscore'
 Rest                     = require './connect/rest'
+TaskQueue                = require './task-queue'
 RepeaterTaskQueue        = require '../lib/repeater-task-queue'
 CartDiscountService      = require './services/cart-discounts'
 CartService              = require './services/carts'
@@ -75,6 +76,11 @@ ALL_SERVICES = [
 # To optimize processing lots of requests all together, e.g.: avoiding connection timeouts, we introduced {TaskQueue}.
 # Every request is internally pushed in a queue which automatically starts resolving promises (requests)
 # and will process concurrently some of them based on the `maxParallel` parameter. You can set this parameter by calling {::setMaxParallel}.
+#
+# ### RepeaterTaskQueue
+# To repeat request on case of some errors. It's possible to use custom repeater by passing it to constructor.
+# Also it's possible to turn off by passing `enableRepeater = false` flag.
+# There is a default repeater if constructor parameter is empty.
 #
 # ```coffeescript
 # client = new SphereClient # a TaskQueue is internally initialized at this point with maxParallel of 20
@@ -221,7 +227,10 @@ class SphereClient
   # options - An {Object} to configure the client
   constructor: (options = {}) ->
     # Private: instance of a {TaskQueue}
-    @_task = options.task or new RepeaterTaskQueue {}, { attempts: 50, timeout: 200, timeoutType: 'v' }
+    if options.enableRepeater? and !options.enableRepeater
+      @_task = new TaskQueue
+    else
+      @_task = options.task or new RepeaterTaskQueue {}, { attempts: 50, timeout: 200, timeoutType: 'v' }
 
     # Private: instance of a {Rest}
     @_rest = options.rest or new Rest _.defaults options, {user_agent: 'sphere-node-sdk'}
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -1,6 +1,6 @@
-_ = require 'underscore'
-Rest = require './connect/rest'
-TaskQueue = require './task-queue'
+_                        = require 'underscore'
+Rest                     = require './connect/rest'
+RepeaterTaskQueue        = require '../lib/repeater-task-queue'
 CartDiscountService      = require './services/cart-discounts'
 CartService              = require './services/carts'
 CategoryService          = require './services/categories'
@@ -220,9 +220,8 @@ class SphereClient
   #
   # options - An {Object} to configure the client
   constructor: (options = {}) ->
-
     # Private: instance of a {TaskQueue}
-    @_task = options.task or new TaskQueue
+    @_task = options.task or new RepeaterTaskQueue {}, { attempts: 50, timeout: 200, timeoutType: 'v' }
 
     # Private: instance of a {Rest}
     @_rest = options.rest or new Rest _.defaults options, {user_agent: 'sphere-node-sdk'}
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -230,7 +230,7 @@ class SphereClient
     if options.enableRepeater? and !options.enableRepeater
       @_task = new TaskQueue
     else
-      @_task = options.task or new RepeaterTaskQueue {}, { attempts: 50, timeout: 200, timeoutType: 'v' }
+      @_task = options.task or new RepeaterTaskQueue {}, {}
 
     # Private: instance of a {Rest}
     @_rest = options.rest or new Rest _.defaults options, {user_agent: 'sphere-node-sdk'}
--- a/src/coffee/repeater-task-queue.coffee
+++ b/src/coffee/repeater-task-queue.coffee
@@ -1,0 +1,73 @@
+{ Repeater } = require 'sphere-node-utils'
+Promise = require 'bluebird'
+util = require 'util'
+TaskQueue = require './task-queue'
+
+# Response messages which will be handled by `RepeaterTaskQueue`
+retryKeywords = [
+  'ETIMEDOUT'
+  'socket hang up'
+  'write EPROTO'
+  'Retry later'
+  'I am the LoadBalancer of'
+  'Gateway Timeout'
+  'Bad Gateway'
+  'EAI_AGAIN'
+  'ESOCKETTIMEDOUT'
+  'Oops. This shouldn\'t happen'
+  'InternalServerError: Undefined'
+  'Temporary overloading'
+  'read ECONNRESET'
+  'getaddrinfo ENOTFOUND'
+  'Cannot commit on stream id'
+]
+
+# Public: A `RepeaterTaskQueue` adds request repeater on particular response errors
+#
+# `RepeaterTaskQueue` receives two objects as parameter.
+# First object overriding `maxParallel` value of `TaskQueue`
+# Second object contains information about the count of attempts, timeout and timeout type.
+# There are 2 types of `timeoutType`:
+# - `c`: constant delay
+# - `v`: variable delay (grows with attempts count with a random component)
+#
+#
+#  Examples
+#
+#   task = new RepeaterTaskQueue { maxParallel: 30 }, { attempts: 50, timeout: 200, timeoutType: 'v' }
+class RepeaterTaskQueue extends TaskQueue
+
+
+  constructor: (options, repeaterOptions) ->
+    super options
+    @repeaterOptions = repeaterOptions
+
+
+  _startTask: (task) =>
+    @_activeCount += 1
+    repeater = new Repeater(@repeaterOptions)
+    toRepeat = repeater.execute task.fn, (err) =>
+      if @_shouldRetry(err)
+        return Promise.resolve()
+      else
+        Promise.reject err
+    toRepeat.then (res) ->
+      task.resolve res
+      return
+    .catch (err) ->
+      task.reject err
+      return
+    .finally =>
+      @_activeCount -= 1
+      @_maybeExecute()
+    .done()
+
+
+  _shouldRetry: (error) ->
+    if error and (error.code and util.inspect(error.code).startsWith('5') or error.statusCode and util.inspect(error.statusCode).startsWith('5'))
+      return true
+    retryKeywords.find (str) ->
+      util.inspect(error, depth: null).toUpperCase().indexOf(str.toUpperCase()) > -1
+      return
+
+module.exports = RepeaterTaskQueue
--- a/src/coffee/repeater-task-queue.coffee
+++ b/src/coffee/repeater-task-queue.coffee
@@ -2,6 +2,7 @@
 Promise = require 'bluebird'
 util = require 'util'
 TaskQueue = require './task-queue'
+_ = require 'underscore'
 
 # Response messages which will be handled by `RepeaterTaskQueue`
 retryKeywords = [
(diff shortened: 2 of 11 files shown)


LEQADA:
@Siilwyn @hisabimbola please review

src/coffee/client.coffee
```diff
@@ -220,9 +220,8 @@ class SphereClient
   #
   # options - An {Object} to configure the client
   constructor: (options = {}) ->
-
     # Private: instance of a {TaskQueue}
-    @_task = options.task or new TaskQueue
+    @_task = options.task or new RepeaterTaskQueue {}, { attempts: 50, timeout: 200, timeoutType: 'v' }
```
LEQADA: Actually 50 attempts is not so bad because we use [Exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) approach for timeout.
A would create disableRepeater flag to turn repeater off. Is it ok?

src/spec/repeater-task-queue.spec.coffee
```diff
@@ -0,0 +1,75 @@
...
+
+  it 'should finally resolve after three tries', (done) ->
+
+    callsMap = {
+      0: { statusCode: 500, message: 'ETIMEDOUT' }
+      1: { statusCode: 500, message: 'ETIMEDOUT' }
+      2: { statusCode: 200, message: 'success' }
+    }
+    callCount = 0
+    spyOn(@client._rest, 'GET').andCallFake (resource, callback) ->
```
LEQADA: Actually there was sinon stubs before but I didn't want to introduce new dependency only for 2-3 tests.

src/coffee/client.coffee
Siilwyn: Prefer letting them customise the number so there will be one option/flag: to modify the amount of retries and to disable it completely. So setting that option to 0 would disable it.

src/spec/repeater-task-queue.spec.coffee
Siilwyn: I would totally agree if we're talking about a non-dev dependency. I'm okay with the current code on the other hand adding one development dependency doesn't have many cons...

src/coffee/client.coffee
LEQADA: You can pass to client constructor your own Repeater to customise `attempts`, `timeout` and `timeoutType`.

hisabimbola: I do not get what you mean by _Exponentials backoff_ and how it explains the fact that we want to do 50 retries on a request in 200 ms.

hisabimbola: > You can pass to client constructor your own Repeater to customise attempts, timeout and timeoutType.

What @Siilwyn and I are trying to say is that should be possible to customise this values.

Value should not be hard-coded.

src/spec/repeater-task-queue.spec.coffee
hisabimbola: It's a devDependency and jasmine is not know for stubbing. The cons are not much but I have no strong opinion on this, so fine with me this way.

src/coffee/client.coffee
LEQADA: @hisabimbola I think that there should be some default values because a lot of projects already uses this SDK and we cannot force them to change Client creation code. 
But as I said you can pass there your own repeater.

hisabimbola: I am not saying we should not have default values at all. I am saying the user should be able to customize this value.

Similar logic done [here](https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/client.coffee#L228)

 and if not passed in by the user when instantiating the class, we use the defaults to instantiate the class.

LEQADA: @hisabimbola please have a look to the last commit

src/coffee/repeater-task-queue.coffee
```diff
@@ -0,0 +1,73 @@
...
+  'getaddrinfo ENOTFOUND'
+  'Cannot commit on stream id'
+]
+
+# Public: A `RepeaterTaskQueue` adds request repeater on particular response errors
+#
+# `RepeaterTaskQueue` receives two objects as parameter.
+# First object overriding `maxParallel` value of `TaskQueue`
+# Second object contains information about the count of attempts, timeout and timeout type.
+# There are 2 types of `timeoutType`:
```
PhilippSpo: If there are only 2 types, could we also make it a boolean to reduce cognitive load on the developers using this?

butenkor: @PhilippSpo It would require breaking change [here](https://github.com/sphereio/sphere-node-utils/blob/master/src/coffee/helpers/repeater.coffee#L28)

PhilippSpo: Sorry I thought this was new code. Nevermind then. 👍

src/coffee/client.coffee
```diff
@@ -220,9 +226,11 @@ class SphereClient
   #
   # options - An {Object} to configure the client
   constructor: (options = {}) ->
-
     # Private: instance of a {TaskQueue}
-    @_task = options.task or new TaskQueue
+    if options.enableRepeater? and !options.enableRepeater
+      @_task = new TaskQueue
```
emmenko: Question: do we really need to keep 2 task queues? Can't we just have e.g: one which can optionally work as repeater but by default is a "normal" task queue?

src/coffee/repeater-task-queue.coffee
```diff
@@ -0,0 +1,73 @@
+{ Repeater } = require 'sphere-node-utils'
+Promise = require 'bluebird'
+util = require 'util'
+TaskQueue = require './task-queue'
+
+# Response messages which will be handled by `RepeaterTaskQueue`
+retryKeywords = [
```
emmenko: Hmm, this looks very fragile. Usually you retry based on the status code. Having hard coded pieces of strings is really not flexible if e.g.: the message changes.
Can we get rid of this list?

src/coffee/repeater-task-queue.coffee
```diff
@@ -0,0 +1,73 @@
...
+      @_maybeExecute()
+    .done()
+
+
+  _shouldRetry: (error) ->
+    if error and (error.code and util.inspect(error.code).startsWith('5') or error.statusCode and util.inspect(error.statusCode).startsWith('5'))
+      return true
+    retryKeywords.find (str) ->
+      util.inspect(error, depth: null).toUpperCase().indexOf(str.toUpperCase()) > -1
+      return
```
emmenko: I don't like this part (see comment above)

src/spec/client/client.spec.coffee
```diff
@@ -77,6 +77,26 @@ describe 'SphereClient', ->
     expect(=> @client.setMaxParallel(0)).toThrow new Error 'MaxParallel must be a number between 1 and 100'
     expect(=> @client.setMaxParallel(101)).toThrow new Error 'MaxParallel must be a number between 1 and 100'
 
+  it 'should not repeat request on error',(done) ->
```
emmenko: "... if repeater is disabled"

src/spec/connect/oauth2.spec.coffee
```diff
@@ -10,7 +10,7 @@ describe 'OAuth2', ->
     expect(oa._options.host).toBe 'auth.sphere.io'
     expect(oa._options.protocol).toBe 'https'
     expect(oa._options.accessTokenUrl).toBe '/oauth/token'
-    expect(oa._options.timeout).toBe 20000
+    expect(oa._options.timeout).toBe 60000
```
emmenko: Why??

src/spec/integration/products-sync.spec.coffee
```diff
@@ -248,7 +248,7 @@ describe 'Integration Products Sync', ->
       expect(updated.variants[1].prices[0].country).toBe 'FR'
       done()
     .catch (error) -> done(_.prettify(error))
-  , 10000 # 10sec
+  , 30000 # 30sec
```
emmenko: Why did you change all of those timeouts?

src/coffee/repeater-task-queue.coffee
emmenko: Also, we should make it configurable in which cases the request should be retried.
For example by having a "matching" function that needs to implemented from the user. That way we don't make any assumptions which requests should be retried.

src/spec/connect/oauth2.spec.coffee
LEQADA: Because sometimes it fails on my machine with old timeout

src/spec/integration/products-sync.spec.coffee
LEQADA: Because sometimes it fails on my machine with old timeout

src/coffee/repeater-task-queue.coffee
LEQADA: For now it's the best way to do that. Because sometimes there are no error codes at all. I think we can change it in future when we will have more information about this errors.

src/coffee/client.coffee
LEQADA: It's possible I think. But now it's easier to read and understand.

src/coffee/repeater-task-queue.coffee
dhettler: @emmenko Not all of the errors in the list return a proper error code. E.g. in some cases we got
```
"code": "EAI_AGAIN",
"errno": "EAI_AGAIN",
}
```
and ENOTFOUND doesn't even contain these properties.

src/spec/client/client.spec.coffee
LEQADA: Done

src/coffee/repeater-task-queue.coffee
butenkor: @emmenko Some examples where status code is not defined:

```
{
    "error": "{ [HttpError: getaddrinfo EAI_AGAIN api-at-rackspace.sphere.io:443]\n  message: 'getaddrinfo EAI_AGAIN api-at-rackspace.sphere.io:443',\n  body: \n   { statusCode: undefined,\n     message: 'getaddrinfo EAI_AGAIN api-at-rackspace.sphere.io:443',\n     originalRequest: \n      { endpoint: '/channels?where=key%3D%22orderexport%22',\n        options: \n         { config: { project_key: 'xxx' },\n           host: 'api-at-rackspace.sphere.io',\n           protocol: 'https',\n           timeout: 360000,\n           rejectUnauthorized: true,\n           headers: \n            { 'User-Agent': 'xxx-cli-0.13.4',\n              Authorization: 'Bearer **********' },\n           uri: 'https://api-at-rackspace.sphere.io/xxx' } } },\n  name: 'HttpError',\n  code: undefined }",
    "hostname": "xxx",
    "level": 50,
    "msg": "'order-export' - Unexpected error after '0d 0h 0m 5.297s'",
    "name": "xxx-cli-0.13.4",
    "pid": 1,
    "time": "2016-09-13T18:25:09.745Z",
    "v": 0
}
```

```
{
    "error": "Error: read ECONNRESET\n    at exports._errnoException (util.js:907:11)\n    at TCP.onread (net.js:557:26)",
    "hostname": "xxx",
    "level": 50,
    "msg": "Error details: ",
    "name": "xxx-cli-0.12.9",
    "pid": 1,
    "time": "2016-09-06T00:46:01.203Z",
    "v": 0
}
```

We could make this list as part of constructor which might be overridden if defined in options?

src/coffee/repeater-task-queue.coffee
butenkor: @emmenko See comment above (user could optionally define a list of messages to retry). You can always extend ```RepeaterTaskQueue```, override ```_shouldRetry``` and pass it to client instead.

src/coffee/client.coffee
```diff
@@ -220,9 +226,11 @@ class SphereClient
   #
   # options - An {Object} to configure the client
   constructor: (options = {}) ->
-
     # Private: instance of a {TaskQueue}
-    @_task = options.task or new TaskQueue
+    if options.enableRepeater? and !options.enableRepeater
```
PhilippSpo: Hm not sure if this was intended, but this condition requires the `enableRepeater` option to exist and to be falsy. This means that if the `enableRepeater` option is not provided it would go into the else block. This looks like an unwanted side-effect to me.

src/coffee/repeater-task-queue.coffee
```diff
@@ -0,0 +1,79 @@
...
+      task.reject err
+      return
+    .finally =>
+      @_activeCount -= 1
+      @_maybeExecute()
+    .done()
+
+
+  _shouldRetry: (error) ->
+    if error and (error.code and util.inspect(error.code).startsWith('5') or error.statusCode and util.inspect(error.statusCode).startsWith('5'))
```
PhilippSpo: This is a bit confusing to read.
If I understand correctly `_shouldRetry` should return true if:
- the `error.code` or `error.statusCode` is a 5xx
- the `error` contains one of the retry keywords
I also don't think we need to use nodes `util` library here.

```coffee
_shouldRetry: (error) ->
  return error?.code.toString().startsWith('5')
    or error?.statusCode.toString().startsWith('5')
    or @repeaterOptions.retryKeywords.some (keyword) -> JSON.stringify(error).toUpperCase().includes(keyword.toUpperCase())
```

src/coffee/client.coffee
LEQADA: @PhilippSpo, as @butenkor [already mentioned](https://github.com/sphereio/sphere-node-sdk/pull/190#discussion_r92011601), the idea is to turn the repeater on if user of SDK will not send any flag to it. So it's not a side-effect, but a feature.

PhilippSpo: Hm. True. Then I think this is not the right place to have this default fallback. It should be done like [here](https://github.com/sphereio/sphere-node-sdk/blob/master/src/coffee/client.coffee#L228). But this is a general problem in the SDK right now as far as I can see.

src/spec/repeater-task-queue.spec.coffee
```diff
@@ -33,7 +34,7 @@ describe 'RepeaterTaskQueue', ->
 
     @client.products.fetch()
     .then (res) =>
-      expect(@client._rest.GET.calls.length).toEqual 3
+      expect(@client._rest.GET.calls.length).toEqual 4
```
PhilippSpo: Now the title of the test is wrong 🤓

PR #192, opened 2016-12-06
feat(order-sync): add line item state support to order synchronization

Siilwyn:
- [x] code is unit tested

### DIFF
--- a/src/coffee/sync/utils/order.coffee
+++ b/src/coffee/sync/utils/order.coffee
@@ -117,6 +117,40 @@ class OrderUtils extends BaseUtils
       .value()
     _.flatten actions
 
+  # Private: map line items
+  #
+  # diff - {Object} The result of diff from `jsondiffpatch`
+  # old_obj - {Object} The existing order
+  #
+  # Returns {Array} The list of actions, or empty if there are none
+  actionsMapLineItems: (diff, old_obj, new_obj) ->
+    actions = []
+    
+    data = new_obj || old_obj
+    
+    # Note the lineItem.state contains 'from/to' instead of just the data
+
+    if data.lineItems
+      data.lineItems.forEach (lineItem) ->
+        if (lineItem.state)
+          lineItem.state.forEach (state) ->
+            if (state.fromState && state.toState)
+              action = {
+                action: 'transitionLineItemState',
+                lineItemId: lineItem.id,
+                quantity: state.quantity,
+                fromState: state.fromState,
+                toState: state.toState,
+              }
+
+              # Check for optional fields
+              if (state.actualTransitionDate)
+                action.actualTransitionDate = state.actualTransitionDate
+
+              actions.push(action)
+
+    actions
+
 module.exports = OrderUtils
 
 #################
--- a/src/coffee/sync/utils/order.coffee
+++ b/src/coffee/sync/utils/order.coffee
@@ -151,6 +151,40 @@ class OrderUtils extends BaseUtils
 
     actions
 
+  # Private: map custom line items
+  #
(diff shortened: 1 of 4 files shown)


src/coffee/sync/utils/order.coffee
```diff
@@ -117,6 +117,37 @@ class OrderUtils extends BaseUtils
...
+    
+    data = new_obj || old_obj
+    
+    # Note the lineItem.state contains 'from/to' instead of just the data
+
+    if data.lineItems
+      data.lineItems.forEach (lineItem) ->
+        if (lineItem.state)
+          lineItem.state.forEach (state) ->
+            if (state.fromState && state.toState)
```
hisabimbola: Since this is an object and empty object is truthy you might want to check the size or values in the object.

Siilwyn: The API will return an error if the content of one of the fields is incorrect.

src/coffee/sync/utils/order.coffee
```diff
@@ -117,6 +117,37 @@ class OrderUtils extends BaseUtils
...
+              actions.push(
+                {
+                  action: "transitionLineItemState",
+                  lineItemId: lineItem.id,
+                  quantity: state.quantity,
+                  fromState: state.fromState,
+                  toState: state.toState,
+                  actualTransitionDate: state.actualTransitionDate
```
Siilwyn: Sharp! Will check it.

Siilwyn:
I do not understand how the coverage decreased, it looks like coveralls says [this line](https://github.com/sphereio/sphere-node-sdk/blob/support-order-line-items/src/coffee/sync/utils/order.coffee#L147) is not covered but as far as I can see it is being check both resolving to false an true in [these unit tests](https://github.com/sphereio/sphere-node-sdk/blob/support-order-line-items/src/spec/sync/utils/order.spec.coffee#L416-L492).
:"|

LEQADA:
[On this line](https://github.com/sphereio/sphere-node-sdk/blob/support-order-line-items/src/spec/sync/utils/order.spec.coffee#L430)  you put extra comma. I bet that you wanted to add `actualTransitionDate: 'someDateHere'` but forgot :)
That's why it's not cover [this line](https://github.com/sphereio/sphere-node-sdk/blob/support-order-line-items/src/coffee/sync/utils/order.coffee#L147)
And if you'll add that `actualTransitionDate` to `expectedActions` then I think coveralls will stop screaming

Siilwyn:
@LEQADA thank you for the suggestion. That comma is there just for trailing, a habit I got into since our new modules require is for the linter. As for adding it that is what [this test](https://github.com/sphereio/sphere-node-sdk/blob/support-order-line-items/src/spec/sync/utils/order.spec.coffee#L468-L488) is for right? Or am I missing something?

src/spec/integration/orders-sync.spec.coffee
```diff
@@ -212,6 +222,70 @@ describe 'Integration Orders Sync', ->
       done()
     .catch (error) -> done(_.prettify(error))
 
+  it 'should sync line items', (done) ->
+    orderNew = _.deepClone @order
+    
+    orderNew.lineItems[0].state = [
+      {
+        quantity: 1,
+        fromState: {
```
PhilippSpo: Why are you doing this? This means that when I use the SDK to generate update actions for line item states, I have to provide the `fromState` and `toState`. This seems wrong. We can just look ad the "old" order to get the from state and at the "new" order to get the new state, right?

Siilwyn:
After talking with Philipp we have decided to implement this in the module that is updating because this is not really diff'ing the data and thus breaks with the concept of the sync modules in the SDK.

PR #197, opened 2017-01-04
fix(product-sync): long text diff error

lojzatran:
How to reproduce this error:
1) You have a product with STRING attribute that contains invalid JSON. The length of this String is long enough so that jsondiffpatch uses text diff algorithm
2) You want to update this String attribute with the valid JSON. This valid JSON is also long enough.
3) As the result, the valid String attribute is changed into invalid JSON String.

I discovered there's an error inside the algorithm itself. The fastest way is to disable the usage of the algorithm. However, it's not possible to disable it, as it will fall back to default value 60: https://github.com/benjamine/jsondiffpatch/blob/0fe51f884695b02179e7a836a82b4241243d78cb/src/filters/texts.js#L55. So for now I increase it to bigger value to prevent using this algorithm.

- [ ] code is unit tested

### DIFF
--- a/src/coffee/sync/utils/base.coffee
+++ b/src/coffee/sync/utils/base.coffee
@@ -18,7 +18,7 @@ class BaseUtils
       textDiff:
         # if value to diff has a bigger length, a text diffing algorithm is used
         # https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md#text-diffs
-        minLength: 300
+        minLength: 50000
 
   # Internal: Compute the difference between given objects
   diff: (old_obj, new_obj) -> @diffpatcher.diff(old_obj, new_obj)
--- a/src/coffee/sync/utils/base.coffee
+++ b/src/coffee/sync/utils/base.coffee
@@ -17,8 +17,9 @@ class BaseUtils
         includeValueOnMove: false  # value of items moved is not included in deltas
       textDiff:
         # if value to diff has a bigger length, a text diffing algorithm is used
+        # disable the text diffing algorithm by passing an unreachable minimum length
         # https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md#text-diffs
-        minLength: 50000
+        minLength: Infinity
 
   # Internal: Compute the difference between given objects
   diff: (old_obj, new_obj) -> @diffpatcher.diff(old_obj, new_obj)
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "debug": "2.1.x",
-    "jsondiffpatch": "0.1.31",
+    "jsondiffpatch": "0.2.4",
     "request": "2.54.x"
   },
   "peerDependencies": {
--- a/src/spec/sync/product-sync.spec.coffee
+++ b/src/spec/sync/product-sync.spec.coffee
@@ -132,9 +132,8 @@ describe 'ProductSync', ->
 
   describe ':: buildActions', ->
 
-    iit 'should build the action update', ->
+    it 'should build the action update', ->
       update = @sync.buildActions(NEW_PRODUCT, OLD_PRODUCT).getUpdatePayload()
-      console.log 'update is ' + JSON.stringify(update.actions[update.actions.length - 1], null, 2)
       expected_update =
         actions: [
           { action: 'removeVariant', id: 4 }
--- a/src/spec/sync/product-sync.spec.coffee
+++ b/src/spec/sync/product-sync.spec.coffee
@@ -22,6 +22,9 @@ OLD_PRODUCT =
       {id: 'p-3', value: {currencyCode: 'EUR', centAmount: 1100}, country: 'DE'},
       {id: 'p-4', value: {currencyCode: 'EUR', centAmount: 1200}, customerGroup: {id: '984a64de-24a4-42c0-868b-da7abfe1c5f6', typeId: 'customer-group'}}
     ]
+    attributes: [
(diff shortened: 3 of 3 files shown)


butenkor:
@Siilwyn @hisabimbola Please comment or approve

hisabimbola:
@lojzatran, please rebase this branch, I think it's outdated

lojzatran:
@butenkor @hisabimbola thanks for reviewing, I already solved the conflicts. Can you merge it to master and create a release?

src/coffee/sync/utils/base.coffee
```diff
@@ -17,8 +17,9 @@ class BaseUtils
         includeValueOnMove: false  # value of items moved is not included in deltas
       textDiff:
         # if value to diff has a bigger length, a text diffing algorithm is used
+        # disable the text diffing algorithm by passing an unreachable minimum length
         # https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md#text-diffs
-        minLength: 50000
+        minLength: Infinity
```
butenkor: 👍

src/coffee/sync/utils/base.coffee
```diff
@@ -17,8 +17,9 @@ class BaseUtils
         includeValueOnMove: false  # value of items moved is not included in deltas
       textDiff:
         # if value to diff has a bigger length, a text diffing algorithm is used
+        # disable the text diffing algorithm by passing an unreachable minimum length
         # https://github.com/benjamine/jsondiffpatch/blob/master/docs/deltas.md#text-diffs
-        minLength: 300
+        minLength: Infinity
```
emmenko: C'mon, seriously?!?

I still don't understand and see how and what kind of a problem you are having.
I tried to reproduce this problem (in the new version of the sdk, I haven't tried in the 1.x) but I don't see any.

```js
describe('long text diff patch', () => {
  it('should correctly patch it', () => {
    const before = {
      attributes: [
        {
          name: 'textAttribute',
          value: '[{"textAttributeValue":{"fr-CH":"","de-CH":"","it-CH":"","de-DE":"<p><strong>Some random text to make this longer than the value that was in jsondiffpatch.textDiff.minLength = 300. Also this will be badly formatted JSON”</p>","en-GB":"","es-ES":"","fr-FR":""fr-CH":"","fr-FR": "","it-IT": "","nl-NL": "","ru-RU": ""},"testberichte_video": ""}]',
        },
      ],
    }
    const now = {
      attributes: [
        {
          name: 'textAttribute',
          value: '[{"textAttributeValue":{"fr-CH":"","de-CH":"","it-CH":"","de-DE":"<p><strong>Some random text to make this longer than the value that was in jsondiffpatch.textDiff.minLength = 300. This should be now a correctly formatted JSON. However, after jsondiffpatch, it will be changed into a different string”</p>","en-GB":"","es-ES":"","fr-FR":""}}]'
        },
      ],
    }
    const delta = diffpatcher.diff(before, now)
    expect(JSON.parse(
      diffpatcher.getDeltaValue(
        delta.attributes['0'].value,
        before.attributes[0].value,
      )
    )).toEqual([
      {
        textAttributeValue: {
          'fr-CH': '',
          'de-CH': '',
          'it-CH': '',
          'de-DE': '<p><strong>Some random text to make this longer than the value that was in jsondiffpatch.textDiff.minLength = 300. This should be now a correctly formatted JSON. However, after jsondiffpatch, it will be changed into a different string”</p>',
          'en-GB': '',
          'es-ES': '',
          'fr-FR': '',
        },
      },
    ])
  })
})

// Delta
{
  "attributes": {
    "0": {
      "value": [
        "@@ -178,166 +178,165 @@\n 00. \n-Also this will be badly formatted JSON%E2%80%9D%3C/p%3E%22,%22en-GB%22:%22%22,%22es-ES%22:%22%22,%22fr-FR%22:%22%22fr-CH%22:%22%22,%22fr-FR%22: %22%22,%22it-IT%22: %22%22,%22nl-NL%22: %22%22,%22ru-RU%22: %22%22%7D,%22testberichte_video%22: %22%22\n+This should be now a correctly formatted JSON. However, after jsondiffpatch, it will be changed into a different string%E2%80%9D%3C/p%3E%22,%22en-GB%22:%22%22,%22es-ES%22:%22%22,%22fr-FR%22:%22%22%7D\n %7D%5D\n",
        0,
        2
      ]
    },
    "_t": "a"
  }
}
```

lojzatran: @emmenko try to run only the test `'should build the action update'` and remove this Infinity part or put it back to 300. It will use google-diff-match-patch algorithm for longer text, which will produce the error.

lojzatran:
If everything is ok, can you merge and make a new release? I'm not allowed to merge to master. Thank you.

lojzatran:
@emmenko so after discussion with @butenkor I finally understand what you meant in your comment. I checked the code from the new version of the sdk and indeed you're right and my fix is not good. The real problem is not in the algorithm in jsondiffpatch, but in the SDK.

In the code there's this line:
`jsondiffpatch.patch(obj, arr)`
https://github.com/sphereio/sphere-node-sdk/blob/41907599b99d7a9912c4cfe544cd1c1b3bf33930/src/coffee/sync/utils/base.coffee#L45

However, the first argument of `patch()` can only be the original value, which is not really clear from the variable name. When you pass the original value, then the diffs can be applied correctly (since it's a diff from the original value). But when the new value is passed, diffs are not applied correctly and produce wrong result. I had a look at the usages of this method and there are some places that might pass the new value instead of the original one, e.g. https://github.com/sphereio/sphere-node-sdk/blob/41907599b99d7a9912c4cfe544cd1c1b3bf33930/src/coffee/sync/utils/product.coffee#L493

@hisabimbola @Siilwyn I will close this PR, fix the error and create a new PR, please don't merge it now.

PR #198, opened 2017-01-17
fix(product-sync): set attributes before add variants

lojzatran:
Product sync updates needs to set attributes before adding new variants. The problem is when you have setAttributeInAllVariants for same for all attribute and a new variant, you need to change the same for all value for old variants first before you can create the new variant.

- [ ] code is unit tested

### DIFF
--- a/src/coffee/sync/product-sync.coffee
+++ b/src/coffee/sync/product-sync.coffee
@@ -37,12 +37,17 @@ class ProductSync extends BaseSync
     super new_obj, old_obj
 
   _doMapActions: (diff, new_obj, old_obj) ->
+    variantActions = @_mapActionOrNot 'variants', => @_utils.actionsMapVariants(diff, old_obj, new_obj)
+    removeVariantActions = variantActions.filter (action) -> action.action is 'removeVariant'
+    addVariantActions = variantActions.filter (action) -> action.action is 'addVariant'
+
     allActions = []
-    allActions.push @_mapActionOrNot 'variants', => @_utils.actionsMapVariants(diff, old_obj, new_obj)
+    allActions.push removeVariantActions
+    allActions.push @_mapActionOrNot 'attributes', => @_utils.actionsMapAttributes(diff, old_obj, new_obj, @sameForAllAttributeNames)
+    allActions.push addVariantActions
     allActions.push @_mapActionOrNot 'base', => @_utils.actionsMapBase(diff, old_obj)
     allActions.push @_mapActionOrNot 'references', => @_utils.actionsMapReferences(diff, old_obj, new_obj)
     allActions.push @_mapActionOrNot 'prices', => @_utils.actionsMapPrices(diff, old_obj, new_obj)
-    allActions.push @_mapActionOrNot 'attributes', => @_utils.actionsMapAttributes(diff, old_obj, new_obj, @sameForAllAttributeNames)
     allActions.push @_mapActionOrNot 'images', => @_utils.actionsMapImages(diff, old_obj, new_obj)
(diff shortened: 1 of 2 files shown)


src/coffee/sync/product-sync.coffee
```diff
@@ -37,12 +37,17 @@ class ProductSync extends BaseSync
     super new_obj, old_obj
 
   _doMapActions: (diff, new_obj, old_obj) ->
+    variantActions = @_mapActionOrNot 'variants', => @_utils.actionsMapVariants(diff, old_obj, new_obj)
+    removeVariantActions = variantActions.filter (action) -> action.action is 'removeVariant'
+    addVariantActions = variantActions.filter (action) -> action.action is 'addVariant'
```
emmenko: Please document why you are doing this and why it is necessary.

lojzatran: https://github.com/sphereio/sphere-node-sdk/pull/198/commits/862893d3dc24758c1e1836e92edcaa8733ef9221

src/coffee/sync/product-sync.coffee
```diff
@@ -37,12 +37,20 @@ class ProductSync extends BaseSync
     super new_obj, old_obj
 
   _doMapActions: (diff, new_obj, old_obj) ->
+    # New variant can have attributes with new values and this can cause conflict if the product still
+    # has old value of the sameForAll attribute. That's why it's necessary to update the attributes
+    # before adding new variant
+    variantActions = @_mapActionOrNot 'variants', => @_utils.actionsMapVariants(diff, old_obj, new_obj)
```
Siilwyn: Why put these functions in separate variables? Think inlining them is more in line with the rest of the code 'style'.

lojzatran: I think it's a bit more readable. WDYT?

```js
allActions.push variantActions.filter (action) -> action.action is 'removeVariant'
allActions.push @_mapActionOrNot 'attributes', => @_utils.actionsMapAttributes(diff, old_obj, new_obj, @sameForAllAttributeNames)
allActions.push variantActions.filter (action) -> action.action is 'addVariant'
```
vs
```js
allActions.push removeVariantActions
allActions.push @_mapActionOrNot 'attributes', => @_utils.actionsMapAttributes(diff, old_obj, new_obj, @sameForAllAttributeNames)
allActions.push addVariantActions
```

But it's not a problem for me to change it as you suggested.

Siilwyn: Just looking at these three lines the second example is more readable. But now to understand the second example you need to look for the function definition so reading through the code goes up and down.

lojzatran: https://github.com/sphereio/sphere-node-sdk/pull/198/commits/48688f361068da459f78b9f351763f9b59546614

You need to go up and down anyway ↕️ , but maybe to make it more in line with the code style is a good idea. 👍

src/coffee/sync/product-sync.coffee
```diff
@@ -37,12 +37,18 @@ class ProductSync extends BaseSync
     super new_obj, old_obj
 
   _doMapActions: (diff, new_obj, old_obj) ->
+    # New variant can have attributes with new values and this can cause conflict if the product still
+    # has old value of the sameForAll attribute. That's why it's necessary to update the attributes
+    # before adding new variant
```
emmenko: This is still not really clear and does not explain the problem -> solution

lojzatran: What about now❓  https://github.com/sphereio/sphere-node-sdk/pull/198/commits/50df3bbe6d9df6c924484d6056515d7daf71aa53

PR #199, opened 2017-01-18
fix(inventory): change quantity instead of calculating the difference

lojzatran:
The main reason for this change is it's easier to repeat the request in case of concurrent modification error. With the old diff calculation, it's necessary to refetch the inventories and recalculate the diff. With this new change, it's enough to change the version number and repeat the request.

- [ ] code is unit tested

### DIFF
--- a/src/coffee/services/inventory-entries.coffee
+++ b/src/coffee/services/inventory-entries.coffee
@@ -13,7 +13,7 @@ BaseService = require './base'
 #       version: result.body.version
 #       actions: [
 #         {
-#           action: 'addQuantity'
+#           action: 'changeQuantity'
 #           quantity: 10
 #         }
 #       ]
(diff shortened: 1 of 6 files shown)


Siilwyn:
This looks almost too good to be true, less code & better functionality. 👀 
There must be a reason why this wasn't done initially though...
Also please check out what's causing the decreased coverage.

lojzatran:
@Siilwyn I think the reason was when node-sdk was created, there was no `changeQuantity` function in the API :)

Siilwyn:
Ah good one, great idea to change it. 👍 🎉

lojzatran:
@Siilwyn coverage is back to normal, I think there was some inconsistencies in coveralls that causes the coverage decrease.

Siilwyn:
It's still decreased, no?

lojzatran:
@Siilwyn ah ok, I looked at it at the wrong time. I'm not sure why the coverage is decreasing as I removed some codes, so it should increase 😕

PR #200, opened 2017-01-19
Fix/long text diff error

lojzatran:
This PR fixes the wrong value returned by the jsondiffpatch. The bug was that in order to call `jsondiffpatch.patch()`, you need to pass an original value and the diff. However, the code passes the new value and the diff, which causes the jsondiffpatch to return wrong value after patch.

Previous PR that was closed because the solution was not correct:
https://github.com/sphereio/sphere-node-sdk/pull/197

- [ ] code is unit tested

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "debug": "2.1.x",
-    "jsondiffpatch": "0.1.31",
+    "jsondiffpatch": "0.2.4",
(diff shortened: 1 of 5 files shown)


hisabimbola:
Hi @lojzatran Can you draft a release [here](https://github.com/sphereio/sphere-node-sdk/releases) and I will merge it and release

PR #205, opened 2017-02-01
fix(product): don't modify original product when building variant attributes

lojzatran:
The old code deleted the attribute of an object that was part of the original product attribute. So instead of correct format:
```js
{ name: 'attributeName', value: 'attributeValue' }
```
we had:
```js
{ name: 'attributeName' }
```
in the variant attributes of the original product. I fixed it by replacing the old code with the one from the new node SDK.
https://github.com/commercetools/nodejs/blob/ab8edfb41bdc7d429f20554d3d8a45ef251228f8/packages/sync-actions/src/product-actions.js#L286

- [ ] code is unit tested

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -562,8 +562,11 @@ class ProductUtils extends BaseUtils
               return
             deltaValue = @getDeltaValue(value)
             unless deltaValue
-              deltaValue = value[0]
-              delete deltaValue.value
+              # Taken from https://github.com/commercetools/nodejs/blob/ab8edfb41bdc7d429f20554d3d8a45ef251228f8/packages/sync-actions/src/product-actions.js#L286
+              if (value[0] && value[0].name)
+                deltaValue = { name: value[0].name }
+              else
+                deltaValue = undefined
             id = old_variant.id
             setAction = @_buildNewSetAttributeAction(id, deltaValue, sameForAllAttributeNames)
             actions.push setAction if setAction
--- a/src/spec/sync/utils/product.spec.coffee
+++ b/src/spec/sync/utils/product.spec.coffee
@@ -1187,9 +1187,15 @@ describe 'ProductUtils', ->
       cloneOfOriginalProduct = _.deepClone(originalProduct)
 
       delta = @utils.diff originalProduct, newProduct
-      @utils.actionsMapAttributes delta, originalProduct, newProduct
+      update = @utils.actionsMapAttributes delta, originalProduct, newProduct
 
+      expected_update =
+        [
+          { action: 'setAttribute', variantId: 1, name: 'testAttribute1', value: false },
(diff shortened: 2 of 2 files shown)


src/spec/sync/utils/product.spec.coffee
```diff
@@ -1151,6 +1151,46 @@ describe 'ProductUtils', ->
         ]
       expect(update).toEqual expected_update
 
+    it 'should not modify original attributes', ->
```
butenkor: I am confused that original product is used as a base for attribute value and not new product

src/spec/sync/utils/product.spec.coffee
```diff
@@ -1151,6 +1151,46 @@ describe 'ProductUtils', ->
...
+            {
+              name: 'testAttribute2',
+              value: 'testValue'
+            }
+          ]
+
+      cloneOfOriginalProduct = _.deepClone(originalProduct)
+
+      delta = @utils.diff originalProduct, newProduct
+      @utils.actionsMapAttributes delta, originalProduct, newProduct
```
butenkor: should not you check that it adds new attributes (testAttribute1) to make sure that in this edge case it works also?

src/spec/sync/utils/product.spec.coffee
lojzatran: In this particular case original product is used because in the new product, the old attribute is not present anymore and therefore needs to be removed from the original product.

src/spec/sync/utils/product.spec.coffee
lojzatran: https://github.com/sphereio/sphere-node-sdk/pull/205/commits/ceab3833e361dbb958ba5f53db1b820bcdb5f1b9

src/coffee/sync/utils/product.coffee
```diff
@@ -562,8 +562,11 @@ class ProductUtils extends BaseUtils
               return
             deltaValue = @getDeltaValue(value)
             unless deltaValue
-              deltaValue = value[0]
-              delete deltaValue.value
+              # Taken from https://github.com/commercetools/nodejs/blob/ab8edfb41bdc7d429f20554d3d8a45ef251228f8/packages/sync-actions/src/product-actions.js#L286
```
Siilwyn: I don't see the added value of this comment.

lojzatran: I just wanted to say that I did not create the code by myself, but got it from a trustworthy resource :) And in case you'll find an error here in the future, you should also check the link.

src/coffee/sync/utils/product.coffee
```diff
@@ -562,8 +562,11 @@ class ProductUtils extends BaseUtils
...
+              if (value[0] && value[0].name)
```
hisabimbola: Easier if `value[0]` is refactored to a var

lojzatran: I didn't want to differ from the linked resource as it's easier to compare like this.

PR #206, opened 2017-02-02
fix(product): dont create setAttribute action when the product attribute did not change

lojzatran:
When the new product attribute contains same elements as the old product attribute, but only in different order, the update action is created and sent to the API. However, since nothing was changed, an error is returned from the API:

BadRequest: A duplicate combination of the variant values (sku, key, images, prices, attributes) exists.

Solution: check if the product attribute has really changed or the elements of the attribute are just in a different order.

- [ ] code is unit tested

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -470,6 +470,14 @@ class ProductUtils extends BaseUtils
         imageUrl: image.url
     action
 
+  _isExistingAttribute: (oldAttribute, newAttribute) ->
+    oldValue = oldAttribute.value
+    newValue = newAttribute.value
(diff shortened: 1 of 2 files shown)


lojzatran:
I just discovered that my fix is not correct, so please don't merge this yet. I will need to change the PR.

lojzatran:
After some clarifications with the product team it seems like my PR is correct and there's nothing to change here. There is an inconsistent behaviour between setAttribute and setAttributeInAllVariants actions. setAttribute throws an error in case only order in a set is changed, but setAttributeInAllVariants does not throw and actually changes the order as well.

In either case, the order of elements does not matter, and we should not send update action if nothing is changed.

PR #208, opened 2017-02-07
feat(product-projections): add field to toggle markMatchingVariants

### LINKED ISSUE #207: Enable markMarchingVariants
### Expected behavior
Via the SDK, we need the option to set ```markMatchingVariant``` to ```true``` to enable this behaviour. (http://dev.commercetools.com/http-api-projects-products-search.html#matchingVariant)

### Actual behavior
At the moment, the parameter is not supported in the search command.

hisabimbola:
Fixes #207

- [ ] code is unit tested

### DIFF
--- a/src/coffee/services/product-projections.coffee
+++ b/src/coffee/services/product-projections.coffee
@@ -6,7 +6,8 @@ BaseService = require './base'
 # A [`ProductProjection`](http://dev.sphere.io/http-api-projects-products.html#product-projection) is a representation
 # of a the `current` or `staged` version of a product (**only GET requests**).
 #
-# _Products are the sellable goods in an e-commerce project on SPHERE.IO. This document explains some design concepts of products on SPHERE.IO and describes the available HTTP APIs for working with them._
+# _Products are the sellable goods in an e-commerce project on SPHERE.IO.
+# This document explains some design concepts of products on SPHERE.IO and describes the available HTTP APIs for working with them._
 #
 # Examples
 #
@@ -25,6 +26,7 @@ class ProductProjectionService extends BaseService
     super()
     _.extend @_params.query,
       fuzzy: false
+      markMatchingVariants: false
       staged: false
       filter: []
       filterByQuery: []
@@ -33,7 +35,7 @@ class ProductProjectionService extends BaseService
       searchKeywords: []
 
     @_params.encoded = @_params.encoded.concat(['filter', 'filter.query', 'filter.facets', 'facets', 'searchKeywords'])
-    @_params.plain = @_params.plain.concat(['staged', 'fuzzy'])
+    @_params.plain = @_params.plain.concat(['staged', 'fuzzy', 'markMatchingVariants'])
 
   # Public: Define whether to query for staged or current product projection.
   #
@@ -74,9 +76,8 @@ class ProductProjectionService extends BaseService
   # Examples
   #
   #   service = client.productProjections
-  #   service.text('Sapphure', 'en').markMatchingVariants(true).search()
-  markMatchingVariants: (value) ->
-    return this unless _.isBoolean(value)
+  #   service.text('Sapphure', 'en').markMatchingVariants().search()
+  markMatchingVariants: (value = true) ->
     @_params.query.markMatchingVariants = value
     debug 'Setting markMatchingVariants: %s', value
     this
@@ -245,6 +246,7 @@ class ProductProjectionService extends BaseService
     {staged, fuzzy, text, filter, filterByQuery, filterByFacets, facet, searchKeywords, markMatchingVariants} = _.defaults @_params.query,
       fuzzy: false
       staged: false
+      markMatchingVariants: false
       filter: []
       filterByQuery: []
       filterByFacets: []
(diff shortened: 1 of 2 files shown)


src/spec/client/services/product-projections.spec.coffee
```diff
@@ -65,6 +65,18 @@ describe 'ProductProjectionService', ->
...
     expect(@service.fuzzy()._queryString()).toBe 'fuzzy=true'
 
+  it 'should query for markMatchingVariants when set to true', ->
+    expect(@service.markMatchingVariants(true)._queryString()).toBe 'markMatchingVariants=true'
+
+  it 'should query for markMatchingVariants when set to false', ->
+    expect(@service.markMatchingVariants(false)._queryString()).toBe 'markMatchingVariants=false'
+
+  it 'should query for not set anything when no value is passed to markMatchingVariants', ->
+    expect(@service.markMatchingVariants()._queryString()).toBe ''
```
butenkor: should it not default to ```'markMatchingVariants=false'```?

src/spec/client/services/product-projections.spec.coffee
```diff
@@ -98,10 +110,11 @@ describe 'ProductProjectionService', ->
       .filter('foo:bar')
       .filterByQuery('foo:bar')
       .filterByFacets('foo:bar')
+      .markMatchingVariants(true)
```
butenkor: Query for the test above does not use search [endpoint](https://github.com/sphereio/sphere-node-sdk/blob/ae72a3718dfda20adb45c91407eba7a7b653fea1/src/coffee/services/product-projections.coffee#L306) but markMatchingVariants is exclusive feature of search which is not available over product-projections endpoint.

src/coffee/services/product-projections.coffee
```diff
@@ -65,6 +65,22 @@ class ProductProjectionService extends BaseService
...
+  # markMatchingVariants - {Boolean} `true` to do a mark product variants, `false` not mark matching product variants
+  #
+  # Returns a chained instance of `this` class
+  #
+  # Examples
+  #
+  #   service = client.productProjections
+  #   service.text('Sapphure', 'en').markMatchingVariants(true).search()
+  markMatchingVariants: (value) ->
+    return this unless _.isBoolean(value)
```
Siilwyn: Could you explain this?

src/spec/client/services/product-projections.spec.coffee
hisabimbola: I did not add any default to the `markMatchingVariants` method, when it's called you need to explicitly pass a boolean to specify whether it should be false or not.

src/spec/client/services/product-projections.spec.coffee
hisabimbola: Calling `asSearch()` adds the search endpoint

src/coffee/services/product-projections.coffee
hisabimbola: Returns `this` if the value passed in is not a `value`. This enforces the usage of just `true` or `false`

Siilwyn: I stil don't get why you would return `this`.

butenkor: I guess it is required if you want to chain function calls like that: ```service.markMatchingVariants(true).limit(10).search()```

hisabimbola: Consistency and chaining

src/spec/client/services/product-projections.spec.coffee
butenkor: Maybe @yanns knows if it should be disabled by default?

src/spec/client/services/product-projections.spec.coffee
butenkor: I just wanted to point out that only when you do ```asSearch``` only than it makes sense to set markMatchingVariants.

src/spec/client/services/product-projections.spec.coffee
yanns: @agourlay can you take care of that?

agourlay: At the API level ```markMatchingVariants``` is enabled by default.

As we want to disable it by default in the future at the API, I believe it is a good idea to start the change from the clients.

Therefore I vote for disabling by default in the SDK if possible in order to push actively the migration.

Check this for some context https://sphere.atlassian.net/wiki/display/S/Search+matching+variants+disabled+by+default

src/coffee/services/product-projections.coffee
Siilwyn: So if you give it invalid input it just continues?

src/spec/client/services/product-projections.spec.coffee
hisabimbola: Updated and set `markMatchingVariants` to false by default. d9ec17d

hisabimbola:
Closing the PR, this feature will be implemented in the new SDK. You can track progress [here](https://github.com/commercetools/nodejs/issues/100)

PR #212, opened 2017-02-10
chore(package.json): move underscore-mixins to devDependencies

adnasa:

This has been discussed with @Siilwyn


hisabimbola:
I don't quite understand the reason for the move

Siilwyn:
Should be in dependencies and not dev dependencies. :P

Issue #219, opened 2017-03-10
Add subscription endpoint


lojzatran:
### Expected behavior
Add new subscription class to communicate with subscription endpoint in CTP.

### Actual behavior
There's no subscription endpoint.

PR #220, opened 2017-03-10
Feature/subscriptions

lojzatran:
As discussed in https://github.com/sphereio/sphere-node-sdk/pull/159, I added tests and improved the subscription endpoint so it's ready for production use.

Fix https://github.com/sphereio/sphere-node-sdk/issues/219

- [ ] code is unit tested

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,11 @@
 # Contributing
 
+## Setting Up a Local Copy
+1. Clone the repo with git clone git@github.com:sphereio/sphere-node-sdk.git
+2. Run `npm install` in the root nodejs folder. This will ensure that all package dependencies are properly installed / linked.
+3. Run `create_config.sh` and provide your SPHERE.IO and IRON.IO MQ credentials as environmental variables
+4. To run all packages tests run `grunt test`    
+
 ## Committing
 In order to keep commit messages consistent and to be able to generate changelogs automatically we use `commitizen` with `cz-conventional-changelog`. You can read more about it [here](https://commitizen.github.io/cz-cli/). We encourage you to watch this [video tutorial](https://egghead.io/lessons/javascript-how-to-write-a-javascript-library-committing-a-new-feature-with-commitizen) on how to commit a new feature with commitizen.  
 So like shown in the video above, to commit changes you simply enter `npm run commit` and follow the guide.
--- a/create_config.sh
+++ b/create_config.sh
@@ -8,3 +8,12 @@ exports.config = {
   project_key: "${SPHERE_PROJECT_KEY}"
 }
 EOF
+
+cat > "test-config.js" << EOF
+/* IRON.IO MQ credentials */
+exports.config = {
+  iron_mq_url: "${IRON_MQ_URL}",
+  iron_project_id: "${IRON_PROJECT_ID}",
+  iron_token: "${IRON_TOKEN}"
+}
+EOF
--- a/create_config.sh
+++ b/create_config.sh
@@ -5,12 +5,11 @@ cat > "config.js" << EOF
 exports.config = {
   client_id: "${SPHERE_CLIENT_ID}",
   client_secret: "${SPHERE_CLIENT_SECRET}",
-  project_key: "${SPHERE_PROJECT_KEY}"
-}
-/* IRON.IO MQ credentials */
-exports.ironConfig = {
-  iron_mq_url: "${IRON_MQ_URL}",
-  iron_project_id: "${IRON_PROJECT_ID}",
-  iron_token: "${IRON_TOKEN}"
+  project_key: "${SPHERE_PROJECT_KEY}",
+  iron: {
+    mq_url: "${IRON_MQ_URL}",
+    project_id: "${IRON_PROJECT_ID}",
+    token: "${IRON_TOKEN}",
+  },
 }
 EOF
--- a/create_config.sh
+++ b/create_config.sh
@@ -7,11 +7,8 @@ exports.config = {
   client_secret: "${SPHERE_CLIENT_SECRET}",
   project_key: "${SPHERE_PROJECT_KEY}"
 }
-EOF
-
-cat > "test-config.js" << EOF
 /* IRON.IO MQ credentials */
-exports.config = {
+exports.ironConfig = {
   iron_mq_url: "${IRON_MQ_URL}",
   iron_project_id: "${IRON_PROJECT_ID}",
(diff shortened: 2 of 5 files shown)


create_config.sh
```diff
@@ -7,4 +7,13 @@ exports.config = {
   client_secret: "${SPHERE_CLIENT_SECRET}",
   project_key: "${SPHERE_PROJECT_KEY}"
 }
+EOF
+
+cat > "test-config.js" << EOF
```
hisabimbola: Why is this another file??

config.js is mainly for development(testing)

src/spec/integration/subscriptions.spec.coffee
```diff
@@ -0,0 +1,71 @@
...
+Config = require('../../config').config
+testConfig = require('../../test-config').config
+Promise = require 'bluebird'
+debug = require('debug')('spec-integration:products')
+
+newSubscription =
+  key: 'testKey'
+  destination:
+    type: 'IronMQ'
+    uri: "https://#{testConfig.iron_mq_url}/3/projects/#{testConfig.iron_project_id}/queues/test-queue/webhook?oauth=#{testConfig.iron_token}"
```
hisabimbola: _https_ should not be hard coded. and it's much more better that the user provides the full uri. There is no assurance that the format you using to construct the uri will remain constant.

src/spec/integration/subscriptions.spec.coffee
```diff
@@ -0,0 +1,71 @@
...
+    .then (response) =>
+      expect(response.statusCode).toBe 201
+      expect(response.body.id).toBeDefined()
+      done()
+  , 20000
+
+  it 'should update message subscriptions', (done) ->
+    @client.subscriptions.save(newSubscription)
+    .then (response) =>
+      @client.subscriptions.byId(response.body.id).update(
```
hisabimbola: The syncAction class should be used for this...

hisabimbola:
Also _contributing.md_ should be updated that iron.io credentials is needed for the tests to work.

So anybody working on the project needs to provide that...

src/spec/integration/subscriptions.spec.coffee
lojzatran: Do you think it's necessary to create a syncAction class? I don't think that subscriptions will be changed that often. Most of the time you just create a subscription and that's it.

hisabimbola: Sadly that is your use *current* case, it will be too risky to assume that for others too..

lojzatran: So isn't it better for someone who will need it, to create an issue and a PR afterwards? I think it's a different feature. Doesn't look like it's a rule that when we create a new endpoint, we need to create a syncAction class as well.

lojzatran:
https://github.com/sphereio/sphere-node-sdk/pull/220/commits/7c0c38cd43ca74c449c76c2ea40485be5daa2b34

create_config.sh
lojzatran: https://github.com/sphereio/sphere-node-sdk/pull/220/commits/7c0c38cd43ca74c449c76c2ea40485be5daa2b34

src/spec/integration/subscriptions.spec.coffee
lojzatran: https://github.com/sphereio/sphere-node-sdk/pull/220/commits/7c0c38cd43ca74c449c76c2ea40485be5daa2b34

src/spec/integration/subscriptions.spec.coffee
butenkor: @hisabimbola i think you mix sync functionality (which is optional convenience on top of sdk) with pure endpoint integration or you really intend to provide sync actions for every service defined here: https://github.com/sphereio/sphere-node-sdk/tree/master/src/coffee/services?

create_config.sh
```diff
@@ -7,11 +7,8 @@ exports.config = {
   client_secret: "${SPHERE_CLIENT_SECRET}",
   project_key: "${SPHERE_PROJECT_KEY}"
 }
-EOF
-
-cat > "test-config.js" << EOF
 /* IRON.IO MQ credentials */
-exports.config = {
+exports.ironConfig = {
```
hisabimbola: Any reason for putting it in a separate object?

src/spec/integration/subscriptions.spec.coffee
hisabimbola: Okay, we can leave it out..

CONTRIBUTING.md
```diff
@@ -1,5 +1,11 @@
 # Contributing
 
+## Setting Up a Local Copy
```
hisabimbola: 👍

src/spec/integration/subscriptions.spec.coffee
```diff
@@ -0,0 +1,71 @@
...
+    type: 'IronMQ'
+    uri: "#{ironConfig.iron_mq_url}/3/projects/#{ironConfig.iron_project_id}/queues/test-queue/webhook?oauth=#{ironConfig.iron_token}"
+  messages: [
+    resourceTypeId: 'inventory-entry'
+    types: ['InventoryEntryDeleted']
+  ]
+
+describe 'Integration Subscriptions', ->
+
+  beforeEach (done) ->
```
hisabimbola: No need to use `done` for this...

hisabimbola: Also, why not remove before test....

create_config.sh
lojzatran: If I start putting everything in one config object, the object can become too big. Additionally, I can also have object keys with same names.

src/spec/integration/subscriptions.spec.coffee
lojzatran: I don't remove before test because I wanted to be consistent with other tests. They also remove `after` the test. I think it's just extra API call for nothing, right? Each test should start with clean environment and not clean up after previous tests.

lojzatran: I removed `done`: https://github.com/sphereio/sphere-node-sdk/pull/220/commits/eba94d9081d4ebc54edb2f194e6ce559e45432d1

lojzatran:
Hi @hisabimbola can you approve it now or is there sth still missing? Thanks.

hisabimbola:
I tried running on my machine, encountered a missing module.

Working on it now and also improving the config configuration

PR #221, opened 2017-03-22
feat(gzip): allow gzip by default for all requests

### LINKED ISSUE #196: Support gzip compression by default for all requests
### Expected behavior
- All requests should use gzip as default to improve performance

### Actual behavior
- At the moment, no requests sent via the node-sdk are using this, which leads to performance degradation

---

wizzy25:
fixes #196

### DIFF
--- a/src/coffee/client.coffee
+++ b/src/coffee/client.coffee
@@ -152,6 +152,7 @@ ALL_SERVICES = [
 #   stats:
 #     includeHeaders: true
 #     maskSensitiveHeaderData: false
+#   gzipEnable: true # not necessary as gzip is enabled by default. Set to false to disable
 # client.products.fetch()
 # .then (result) ->
 #   # result.statusCode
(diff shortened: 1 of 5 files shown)


butenkor:
@hisabimbola I am just curious - why is it a breaking change and why would you do it optional?

wizzy25:
@hisabimbola please review again

hisabimbola:
@butenkor 
> why is it a breaking change
because the new behaviour differs from the previous implementation
> why would you do it optional?
I don't get what you mean by this, but what this PR did was add gzip flag by default and gave option to turn it off if desired...

butenkor:
How functional behaviour is changed if request data to API was not compressed before and now it is?

Why would anyone want to switch it off if it improves transfer speed?

hisabimbola:
@wizzy25 

Let's update this PR to make it enable by default and deploy this as a new feature not ~breaking change~

wizzy25:
It's presently enabled by default @hisabimbola

yanns:
Thx for pushing this!!
It will help a lot, in impex for example.

wizzy25:
@hisabimbola I think the `package.json` files of the modules and impex should be updated after this is released, and possibly added somewhere (release notes?) in impex

Issue #223, opened 2017-03-29
Improve fetch all method


lojzatran:
### Description
When you doing fetch all, it should fetches all items with minimum requests.

### Current Behavior
Right now fetch all has a limit of 50 items per request and if there are more than 50 items, it needs to fetch multiple times.

### Context
Because of the multiple fetches, there are more queries and the whole fetch all method is slower.

### Possible Solution
Increase the limit number of items to 500, which is the maximum in commercetools platform.

hisabimbola:
Instead of using `fetchAll` you can use the normal `fetch` and specify the limit

hisabimbola:
Also, in place like impex where this module is being used, fetching 500 items by default to the browser might not be too performant..

lojzatran:
@hisabimbola The fix that I try to address here affects only when you do `all().fetch()`. The previous code was fetching 50 items per request, which has the effect of making more requests when you have more than 50 items. I just want to raise it to the maximum 500, which results in less requests in case of `all().fetch()`. This does not affect fetching without `all()`.

hisabimbola:
Okay, that's fine

lojzatran:
Since this can be a breaking change, I will now make the limit parameter configurable and configure it in the modules that need it.

PR #224, opened 2017-03-29
fix(rest): improve fetch all method

lojzatran:
#### Summary
Fixes https://github.com/sphereio/sphere-node-sdk/issues/223

#### Description
Raises the max fetch limit when fetching all items.

### DIFF
--- a/src/coffee/connect/rest.coffee
+++ b/src/coffee/connect/rest.coffee
@@ -193,7 +193,7 @@ class Rest
     throw new Error 'Query limit doesn\'t seem to be 0. This function queries all results, are you sure you want to use this?' if query.limit and query.limit isnt '0'
 
     params = _.extend {}, query,
-      limit: 50 # limit used for batches
+      limit: 500 # limit used for batches
     limit = params.limit
     debug 'PAGED request params: %j', params
 
--- a/src/spec/connect/rest.spec.coffee
+++ b/src/spec/connect/rest.spec.coffee
@@ -211,6 +211,7 @@ describe 'Rest', ->
 
     describe ':: PAGED', ->
 
+      pagedLimit = 500
       idCounter = 0
       uniqueId = (prefix) ->
         id = ++idCounter + ''
@@ -224,15 +225,15 @@ describe 'Rest', ->
           access_token: 'foo'
         @pagedRest = new Rest opts
 
-        offset = -50
-        count = 50
+        offset = -pagedLimit
+        count = pagedLimit
         spyOn(@pagedRest, '_doRequest').andCallFake (options, callback) ->
-          offset += 50
+          offset += pagedLimit
           callback null, {statusCode: 200},
             total: 1004
             count: if offset is 1000 then 4 else count
             offset: offset
-            results: _.map (if offset is 1000 then [1..4] else [1..50]), (i) ->
+            results: _.map (if offset is 1000 then [1..4] else [1..pagedLimit]), (i) ->
               {id: uniqueId("_#{i}") , value: 'foo'}
         spyOn(@pagedRest._oauth, 'getAccessToken')
           .andCallFake (callback) -> callback(null, null, {access_token: 'foo'})
@@ -249,7 +250,7 @@ describe 'Rest', ->
           expect(b.count).toBe 1004
           expect(b.offset).toBe 1000
           expect(b.results.length).toBe 1004
-          expect(@pagedRest._doRequest.calls.length).toBe 21
+          expect(@pagedRest._doRequest.calls.length).toBe 3
           done()
 
       it 'should send PAGED request and preserving query parameters', (done) ->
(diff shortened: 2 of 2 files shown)


src/coffee/connect/rest.coffee
```diff
@@ -193,7 +193,7 @@ class Rest
     throw new Error 'Query limit doesn\'t seem to be 0. This function queries all results, are you sure you want to use this?' if query.limit and query.limit isnt '0'
 
     params = _.extend {}, query,
-      limit: 50 # limit used for batches
+      limit: 500 # limit used for batches
```
hisabimbola: Looking at your comment in the issue, changing this to `500` will affect all limit in the SDK

Since this affects *only* the `all().fetch()` method, you can add the limit params [here](https://github.com/sphereio/sphere-node-sdk/blob/5437331aa292449d6bbeaf7309d6c335e3b1fdf8/src/coffee/services/base.coffee#L321)

lojzatran: Actually this method is used only for `all().fetch()`. There's also a comment at the beginning of this method:

`Public: Fetch all results of a Sphere resource query endpoint in batches of pages using a recursive function`

emmenko: This is not necessary, you can do `all().perPage(500).fetch()`.

emmenko:
This change is not necessary, please close this.

src/coffee/connect/rest.coffee
lojzatran: Yes, I can, but why not to put it here and fetch by 50 instead? This method is by default not optimized, right? When you fetch all, you want to fetch it with as minimum requests as possible.

butenkor:
@emmenko what is the point of fetching 50 items per page if you have to load all items into memory anyway? It makes the process by default slower for anyone who is using it including this module: https://github.com/sphereio/sphere-stock-import/blob/292cfc5e827917eda648b0ca40473e12eb9ea886/src/coffee/stockimport.coffee#L307
So instead of fixing all modules and adding on every .all.fetch() query ```perPage(500)``` we could fix it here instead and adapt dependency for affected modules.

emmenko:
We're not introducing a breaking change now, and this change is not required because you can configure the query. And if you want to use it in some intermediate tools then you should probably make those tools configurable, not changing a default value in the SDK.

I'm not sure also how much slower it is. Do you have numbers to compare?

lojzatran:
@emmenko I don't have exact numbers right now, but we tried to do this change in the stock-sync module and our process was at least 2x faster. However, this process consists of more than just this one module.

I tried to do a inventory fetch request for 50 inventories and for 500 inventories in the Playground. 
For 50 inventories it took around 170ms to fetch.
For 500 inventories it took around 300ms to fetch.

If you have more than 50 inventories, you would need to make 2 requests, which would take already same amount of time as 1x 500 inventory fetch request. So for 500 inventories, you would need 10x fetch requests for 170 ms instead of just 1x 300ms fetch. And as we often have more than 500 inventories to process, the performance can be improved a lot.

I think it's better to fetch everything in one request than do it in many smaller requests (e.g. less latency). Why do you think this is a breaking change? As I wrote before, this is used only when you want to fetch all. When is it better to have many small requests instead of one big?

emmenko:
> I think it's better to fetch everything in one request than do it in many smaller requests (e.g. less latency)

Well yes and no. You're talking about inventories here, which are fairly small entities. If you consider products though, you might run into some problems as you are effectively downloading MBs of data, possibly getting timeouts.

> Why do you think this is a breaking change?

It might not be a breaking change, but we don't know who is using this and how, and we're not sure that just by setting this number higher will simply work seamlessly as before.

> When is it better to have many small requests instead of one big?

I would say in general I prefer to avoid having "long open" connections.

---
But anyway, if you want to take responsibility with that then go ahead 😬

butenkor:
@emmenko JVM SDK does it: https://github.com/commercetools/commercetools-jvm-sdk/blob/master/commercetools-convenience/src/main/java/io/sphere/sdk/queries/QueryExecutionUtils.java#L13

API also officially supports this: https://dev.commercetools.com/http-api.html#limit

If payload is gzipped than even 500 huge products should not be an issue.

yanns:
On the API side, I'd prefer a reasonable default value (50), and that importers/exporters set this value higher when they need.
With a default value of 500, some dev will use it without thinking about it, even if they only show 10 items on one page.

lojzatran:
@yanns I understand your problem, but I'm not sure if it applies in this case. The change I made is for fetching all items, so the only difference to the API is instead of getting more small requests it will get less bigger requests. So if you have 1000 items, in the end it will return you 1000 items anyway, no matter if you set 50 or 500, just the number of requests to the platform is different.
So the question here is is it better for API to have more requests which will return small responses OR is it better to have less bigger requests?

yanns:
@lojzatran it completely depends on the use-case. 
For web applications, it's better to have pagination and smaller requests.
For imports/exports, it's better to have less requests with bigger payloads.

butenkor:
@yanns as @lojzatran mentioned in old and current/new version we load everything what platform has for given query paginated into memory. The questions is if we paginate with 50 items per page and on total result of 1000 items do 20 requests or rather do 4 request before we return results to the caller? We tested it with inventories and from consumer point of view it is at least 2 times faster with 4 request than doing 20 requests.

yanns:
/quote we load everything what platform has for given query paginated into memory.

By "we", do you mean the SDK itself, or the importer/exporter?

butenkor:
SDK

butenkor:
```fetch.all()``` is special as it always loads all it got from API for given query and returns aggregated results. to the caller

yanns:
`all().fetch()` is also quite dangerous. OK to change that.
But I'd prefer focusing on streaming API.

lojzatran:
@yanns if you use it only `all().fetch()`, it's dangerous. But there are cases you use `where().all().fetch()`, which is ok, because you're restricting the results.
So do you think this change is ok then? It should be better for the platform as well if we do less requests, right?

yanns:
@lojzatran difficult to express a general opinion here, because it's depending on the payloads. If you have important products payloads, and also reference expansion, fetching 500 products can be less efficient than 100 for example.

butenkor:
@hajoeichler Then it is ok if we create PR and use perPage(500) [here](https://github.com/sphereio/sphere-stock-import/blob/292cfc5e827917eda648b0ca40473e12eb9ea886/src/coffee/stockimport.coffee#L307), right?

hajoeichler:
@butenkor yes that's fine. Inventory entries are small entities.

PR #227, opened 2017-04-25
MCD-328: User can't add/change inventory quantity to 0

lufego:
#### Summary
At the moment, if you try to add/change a inventory quantity with number "0" the API returns a bad request error

#### Description
This is happening because 0 is falsy, and on the buildBaseAttributesActions function it relies only on the value in order to define what to return

#### Changes made
I have added a variable to check the existence of the value (`hasNow` and `hasBefore`), so 0 wouldn't return false in this case.

#### Todo

- Tests
    - [x] Acceptance

### DIFF
--- a/src/sync/utils/common-actions.js
+++ b/src/sync/utils/common-actions.js
@@ -1,5 +1,6 @@
 import clone from './clone'
 import * as diffpatcher from './diffpatcher'
+import isNil from 'lodash.isnil'
 
 /**
  * Builds actions for simple object properties, given a list of actions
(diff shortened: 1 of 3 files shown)


src/sync/utils/common-actions.js
```diff
@@ -24,18 +24,20 @@ export function buildBaseAttributesActions ({
       const delta = diff[key]
       const before = oldObj[key]
       const now = newObj[key]
+      const hasBefore = oldObj[key] !== null && oldObj[key] !== undefined
```
PhilippSpo: https://lodash.com/docs/4.17.4#isNil ✨

lojzatran:
Hi @lufego, can you explain me why you are fixing for branch `rewrite-2.0`? I thought you use https://github.com/commercetools/nodejs for MC?

adnasa:
@lojzatran not all parts of the MC is using the new `nodejs` yet. We started using when we started implementing the Discounts UI.

Although, you have a point, this can be fixed directly on nodejs, and we change the import in MC.
That should work, right @PhilippSpo @lufego ?

emmenko:
> Although, you have a point, this can be fixed directly on nodejs, and we change the import in MC.

Exactly yes 👍

lufego:
Closing this in favor of https://github.com/commercetools/nodejs/pull/146/files

PR #230, opened 2017-05-22
feat(rest): enable overwriting of page limit

lojzatran:
#### Summary
Enable `limit` parameter to be overwritable from the outside classes.

Fixes: https://github.com/sphereio/sphere-node-sdk/issues/223

### DIFF
--- a/src/spec/connect/rest.spec.coffee
+++ b/src/spec/connect/rest.spec.coffee
@@ -268,18 +268,18 @@ describe 'Rest', ->
           expect(b.results.length).toBe 1004
           expect(@pagedRest._doRequest.calls.length).toBe 21
           expect(@pagedRest._doRequest.calls[0].args[0].uri).toEqual "https://api.sphere.io/#{Config.project_key}/products?" + _.stringifyQuery
+            limit: 50
             where: encodeURIComponent('sku in ("123", "456")')
             expand: 'productType'
             staged: true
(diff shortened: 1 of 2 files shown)


src/spec/connect/rest.spec.coffee
```diff
@@ -314,8 +314,5 @@ describe 'Rest', ->
           expect(b.message).toBe 'Something went wrong'
           done()
 
-      it 'should throw if limit param is not 0', ->
```
hisabimbola: I think we can still leave this test, if limit is 0 we may want to still throw

lojzatran: But this test is about when limit is NOT 0, it will throw an error. We have test for 0 here: https://github.com/sphereio/sphere-node-sdk/pull/230/files/bf0304245c6c899b56c499f3cf2fc8a21a6bdbf7#diff-0525bb4cabce0587175816bbccaefd26R317

hisabimbola: Oh, okay... I get now 👍

hisabimbola:
This is not a "fix" though, looks more like a new feature or enhancement

lojzatran:
OK, I thought it was an error to have 'limit' throwing an error on any error except 0 and still enable you to pass custom value.

butenkor:
@hisabimbola any plans on releasing it?

PR #235, opened 2017-06-23
feat(productsProjection): Add price selection feature to productsProjection endpoint

### LINKED ISSUE #228: Use price selection with Product Projection Search service
### Description

Currently it's not possible to enable Price Selection when using Product Projection Search (or other services). It is possible to send `priceCurrency`, `priceCountry` etc through the REST API, but this SDK doesn't allow sending those at least from what I have figure it out so far.

### Expected Behavior

It should be possible to enable Price Selection using the NodeJS sdk.

junajan:
This PR adds a possibility to turn on a priceSelection feature on productsProjection endpoint.
It is based on the same change in the new SDK - https://github.com/commercetools/nodejs/pull/170

Resolves #228 

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/src/coffee/services/product-projections.coffee
+++ b/src/coffee/services/product-projections.coffee
@@ -344,11 +344,15 @@ class ProductProjectionService extends BaseService
     _.each searchKeywords, (keys) ->
       customQueryString.push "searchKeywords.#{keys.lang}=#{keys.text}"
 
+
     # priceSelection params
-    customQueryString.push "priceCurrency=#{priceCurrency}" if priceCurrency
-    customQueryString.push "priceCountry=#{priceCountry}" if priceCountry
-    customQueryString.push "priceCustomerGroup=#{priceCustomerGroup}" if priceCustomerGroup
-    customQueryString.push "priceChannel=#{priceChannel}" if priceChannel
+    if priceCurrency
+      customQueryString.push "priceCurrency=#{priceCurrency}"
+      customQueryString.push "priceCountry=#{priceCountry}" if priceCountry
+      customQueryString.push "priceCustomerGroup=#{priceCustomerGroup}" if priceCustomerGroup
+      customQueryString.push "priceChannel=#{priceChannel}" if priceChannel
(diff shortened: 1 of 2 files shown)


junajan:
@hisabimbola After this is merged we should publish also new [JS docs](http://sphereio.github.io/sphere-node-sdk/)

src/coffee/services/product-projections.coffee
```diff
@@ -256,6 +344,12 @@ class ProductProjectionService extends BaseService
     _.each searchKeywords, (keys) ->
       customQueryString.push "searchKeywords.#{keys.lang}=#{keys.text}"
 
+    # priceSelection params
+    customQueryString.push "priceCurrency=#{priceCurrency}" if priceCurrency
```
hisabimbola: `priceCurrency` is required to enable price selection, can we throw if `priceCurrency` is not set

adnasa:
Is this really needed?
Shouldn't we stop development, change the maintenance notes and advocate that users move to `nodejs`?

hisabimbola:
Good point @adnasa,

that is the ideal. I agree with that....

hisabimbola:
I think we already promised this feature so we can add this feature especially since there is a PR for it already. But we will try to avoid it moving forward.

adnasa:
What I'm thinking is that once we know that this repo has all the features needed and is in sync with `nodejs/sync-actions` and `nodejs/http`, we can add those notes and change the maintenance status

junajan:
Hello @hisabimbola - please check my update 40312be

adnasa:
To clarify this point:

> What I'm thinking is that once we know that this repo has all the features needed and is in sync with nodejs/sync-actions and nodejs/http, we can add those notes and change the maintenance status

This doesn't mean that we should write any new features here, __unless specifically required__, as for new features in general, we should advocate to have them implemented in `nodejs` first

junajan:
@hisabimbola I do not have rights to merge this PR.. could you do that please?

PR #238, opened 2017-08-09
feat(productSync): Support key and variant key sync action

### LINKED ISSUE #236: Support key attribute in product sync module
I know that this repo has been closed and all new features should go to the [new SDK](https://github.com/commercetools/nodejs/) but there are still some tools which use this SDK and would need some updates so they can support stuff like `product.key` and `variant.key` attributes when [syncing products](https://github.com/sphereio/sphere-node-product-csv-sync/pull/177).

junajan:
#### Summary
Resolves #236 - This PR adds a support for product.key and variant.key sync action.

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -379,6 +379,7 @@ class ProductUtils extends BaseUtils
             updated[k] = diffed_value
 
       # extend with an old object only if value is an object or an array
+      # if the new value is not an array or an object, use just the new value
       if old_obj[key] and (_.isArray(old_obj[key]) or _.isObject(old_obj[key]))
         # extend values of original object with possible new values of the diffed object
         # e.g.:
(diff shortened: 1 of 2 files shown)


src/coffee/sync/utils/product.coffee
```diff
@@ -372,7 +378,8 @@ class ProductUtils extends BaseUtils
             # no idea what this is but lets just use it for updating
             updated[k] = diffed_value
 
-      if old_obj[key]
+      # extend with an old object only if value is an object or an array
```
hisabimbola: can you add the "why" to this comment

junajan: Hello @hisabimbola - I added info here 7fde784 .. We did not have a `key` attribute in the base object before so there were all attributes with a type object and this code was extending new objects with the old ones.

But now we have also attribute 'key' with a type string which can't be extended so I put there this condition which will just use the new version.

PR #249, opened 2018-01-15
fix(repeater): Allow default configuration

junajan:
#### Summary
When [repeaterOptions](https://github.com/sphereio/sphere-node-sdk/blob/247-allow-undefined-configuration/src/coffee/repeater-task-queue.coffee#L41) is not specified:
```js
console.log(_.defaults(undefined, {a: 1})) // -> undefined
```
Which results in error:
```
Cannot read property 'retryKeywords' of undefined
```


junajan:
@wizzy25 - I added a test here c302eb1

PR #251, opened 2018-02-07
feat(sync): utilise priceId for update

### LINKED ISSUE #250: [Sync] Support changePrice
### Description

Currently, the sync actions do not utilise the priceId, so it's not possible to update a price using the priceId

### Expected Behavior

Using the priceId, the [`changePrice`](http://dev.commercetools.com/http-api-projects-products.html#change-price) update action should be triggerable, which can possibly change all fields except the `priceId`
 
### Current Behavior

Due to the non-support of `priceId`, the modifying more than one value in a price triggers a `removePrice` and an `addPrice` update action, which is neither convenient nor necessary

### Context

Initially opened [here](https://github.com/sphereio/sphere-product-import/issues/117)
The sync actions here were probably implemented before the addition of the `priceId` field, hence even the `removePrice` field does not utilise the `priceId`. This should be added

wizzy25:
resolves #250

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -253,8 +253,17 @@ class ProductUtils extends BaseUtils
           actions.push changeAction if changeAction
         else
           delete price.id # delete the id so it doesn't conflict with other actions
+
           # build changePrice update action with price selection
-          if not _.isArray(price) and _.size(price) and not _.has(price, 'country') and not _.has(price, 'channel') and not _.has(price, 'customerGroup') and not _.isArray(price.value?.currencyCode) and old_variant.prices.length > index
+          # at this point, price must be a non-empty object with unchanged price
+          # selection fields to build the chanePrice action
+          if (
+            not _.isArray(price) and _.size(price) and
+            not _.has(price, 'country') and not _.has(price, 'channel') and
+            not _.has(price, 'customerGroup') and
+            not _.isArray(price.value?.currencyCode) and
+            old_variant.prices.length > index
+          )
             changeAction = @_buildChangePriceAction(old_variant, new_variant, index)
             actions.push changeAction if changeAction
           else if _.size(price)
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -249,13 +249,13 @@ class ProductUtils extends BaseUtils
         delete price.discounted # we don't need this for mapping the action
         # if id is unchanged, we initiate a changePrice action with priceId
         if not _.isArray(price) and not price.id and old_variant.prices.length > index
-          changeAction = @_buildChangePriceActionWithPriceId(old_variant, new_variant, index)
+          changeAction = @_buildChangePriceAction(old_variant, new_variant, index)
           actions.push changeAction if changeAction
         else
           delete price.id # delete the id so it doesn't conflict with other actions
           # build changePrice update action with price selection
           if not _.isArray(price) and _.size(price) and not _.has(price, 'country') and not _.has(price, 'channel') and not _.has(price, 'customerGroup') and not _.isArray(price.value?.currencyCode) and old_variant.prices.length > index
-            changeAction = @_buildChangePriceActionWithPriceId(old_variant, new_variant, index)
+            changeAction = @_buildChangePriceAction(old_variant, new_variant, index)
             actions.push changeAction if changeAction
           else if _.size(price)
             removeAction = @_buildRemovePriceAction(old_variant, index)
@@ -408,7 +408,7 @@ class ProductUtils extends BaseUtils
         action = undefined
     action
 
-  _buildChangePriceActionWithPriceId: (old_variant, new_variant, index) ->
+  _buildChangePriceAction: (old_variant, new_variant, index) ->
     new_price = new_variant.prices[index]
     if new_price
       priceId = old_variant.prices[index].id
@@ -419,18 +419,6 @@ class ProductUtils extends BaseUtils
         priceId: priceId
         price: new_price
     action
-
-  # _buildChangePriceAction: (centAmountDiff, variant, index) ->
-  #   price = variant.prices[index]
-  #   if price
-  #     delete price._MATCH_CRITERIA
-  #     delete price.id
-  #     price.value.centAmount = @getDeltaValue(centAmountDiff)
-  #     action =
-  #       action: 'changePrice'
-  #       variantId: variant.id
-  #       price: price
-  #   action
 
   _buildRemovePriceAction: (variant, index) ->
     price = variant.prices[index]
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -19,7 +19,6 @@ class ProductUtils extends BaseUtils
         _.each variant.prices, (price, index) ->
           price._MATCH_CRITERIA = "#{index}"
           delete price.discounted # discount values should not be diffed
-          delete price.id # ids should not be diffed
 
     # Let's compare variants with their SKU, if present.
     # Otherwise let's use the provided id.
@@ -240,6 +239,7 @@ class ProductUtils extends BaseUtils
     actions = []
 
     _mapVariantPrices = (price, key, old_variant, new_variant) =>
+
       if REGEX_NUMBER.test key
         # key is index of new price
         index = key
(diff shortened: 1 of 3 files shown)


src/spec/sync/product-sync.spec.coffee
```diff
@@ -226,6 +226,71 @@ describe 'ProductSync', ->
...
+          }
+        ]
+
+      newProduct =
+        id: '123'
+        masterVariant:
+          sku: 'v1'
+          prices: [
+            {id: 'p-1', value: {currencyCode: 'GBP', centAmount: 555}},
+            {id: 'p-2', value: {currencyCode: 'GBP', centAmount: 245}},
```
butenkor: I am curious what is the business use case that new/updated product variant (supplied by external system) is aware of internal IDs generated by CTP API like price.id?

butenkor: You do not have to rely on price IDs. You can identify/match prices by its scope: currency, country, customer group and channel.

wizzy25: @butenkor There are instances where Prices are exported, modified and imported (with their IDs) back to the platform. Currently, the old prices were removed and the new prices were added. This changes the `priceId` of the price and that's what this issue addresses

The scope fields are optional, and can even be shared by multiple prices in one variant

butenkor: Use case is valid but would fail if you export from production CTP project and want to import into staged project as IDs there will be different. Matching by scope will cover both your case as the one described by me without relying on generated internal IDs.  Even if scope field are optional you can not have prices in API with duplicate scope. So each scope combination (no scope defined is a scope also) is like unique ID which can be used for matching prices on variant level.

wizzy25:
@butenkor I modified the code to also check for a price's uniqueness based on price selection. I also tried to cover as many edge cases as I could

src/coffee/sync/utils/product.coffee
```diff
@@ -248,15 +247,15 @@ class ProductUtils extends BaseUtils
         index = key.substring(1)
       if index
         delete price.discounted # we don't need this for mapping the action
-
         # if id is unchanged, we initiate a changePrice action with priceId
         if not _.isArray(price) and not price.id and old_variant.prices.length > index
-          changeAction = @_buildChangePriceActionWithPriceId(new_variant, index)
+          changeAction = @_buildChangePriceActionWithPriceId(old_variant, new_variant, index)
```
butenkor: `changePrice` actions always require price id so maybe you can skip those suffixes and rename to -> `_buildChangePriceAction`?

Also if we match by scope why do we still check for existence of new price `id` in line 251?

src/coffee/sync/utils/product.coffee
```diff
@@ -248,15 +247,15 @@ class ProductUtils extends BaseUtils
...
           actions.push changeAction if changeAction
         else
           delete price.id # delete the id so it doesn't conflict with other actions
-          if _.size(price) is 1 and _.size(price.value) is 1 and _.has(price.value, 'centAmount')
-            changeAction = @_buildChangePriceAction(price.value.centAmount, old_variant, index)
+          # build changePrice update action with price selection
+          if not _.isArray(price) and _.size(price) and not _.has(price, 'country') and not _.has(price, 'channel') and not _.has(price, 'customerGroup') and not _.isArray(price.value?.currencyCode) and old_variant.prices.length > index
```
butenkor: Can this check be formatted/indented so that it is not in single line and thus better to read? Some comments per check would be nice also as it is not obvious what particular check do and why. For example:
- Why price would be not an array and simultaneously have a array size greater than 0? `_.isArray(price) and _.size(price)`?
- Same question can be applied for `currencyCode`

Why we need two `if` statements in line 251 and 257? Would not simple `old_variant.prices.length > index` be enough for both cases?

src/spec/sync/product-sync.spec.coffee
```diff
@@ -287,10 +287,56 @@ describe 'ProductSync', ->
...
+    it 'should create `changePrice` action if price selection is unique', ->
+      oldProduct =
+        id: '123'
+        version: 1
+        masterVariant:
+          id: 1
+          sku: 'v1'
+          prices: [
+            {id: 'p-1', value: {currencyCode: 'EUR', centAmount: 100}, validUntil: '2019-10-16'},
+            {id: 'p-2', value: {currencyCode: 'EUR', centAmount: 1000}, country: 'DE'},
```
butenkor: Maybe you could add also price with minimum scope also:

`{id: 'p-3', value: {currencyCode: 'EUR', centAmount: 1000}}`

And check if it gets updated also.

wizzy25: Added here: https://github.com/sphereio/sphere-node-sdk/pull/251/commits/ebb0d7498514593062a76580d5fe63dbc8d054c3

wizzy25:
@butenkor what do you think about it now?

src/spec/sync/product-sync.spec.coffee
```diff
@@ -226,6 +226,120 @@ describe 'ProductSync', ->
         version: oldProduct.version
       expect(update).toEqual expected_update
 
+    it 'should create `changePrice` action if new price has id', ->
```
butenkor: If we always match by scope than we do not need a test for price IDs anymore, right?

src/spec/sync/product-sync.spec.coffee
```diff
@@ -226,6 +226,120 @@ describe 'ProductSync', ->
...
+          { action: 'changePrice', priceId: 'p-1', price: {value: {currencyCode: 'GBP', centAmount: 555} }}
+          { action: 'changePrice', priceId: 'p-2', price: {value: {currencyCode: 'GBP', centAmount: 245} }}
+          { action: 'changePrice', priceId: 'p-7', price: {value: {currencyCode: 'USD', centAmount: 4444} }}
+          { action: 'changePrice', priceId: 'p-8', price: {value: {currencyCode: 'USD', centAmount: 5555}, country: 'US' }}
+          { action: 'changePrice', priceId: 'p-10', price: {value: {currencyCode: 'USD', centAmount: 1257}, country: 'US' }}
+        ]
+        version: oldProduct.version
+      expect(update).toEqual expected_update
+
+    it 'should create `changePrice` action if price selection is unique', ->
```
butenkor: Maybe better: `'should create 'changePrice' actions based on price selection'`

src/spec/sync/product-sync.spec.coffee
```diff
@@ -226,6 +226,120 @@ describe 'ProductSync', ->
...
+        ]
+
+      newProduct =
+        id: '123'
+        masterVariant:
+          sku: 'v1'
+          prices: [
+            {value: {currencyCode: 'EUR', centAmount: 555}, validUntil: '2020-12-14'},
+            {value: {currencyCode: 'EUR', centAmount: 245}, country: 'DE'},
+            {value: {currencyCode: 'GBP', centAmount: 2300}}
```
butenkor: Thanks for adding a test. Could you please also mix the order of price objects within prices array of `newProduct.masterVariant` as also change order of properties of price object properties itself? Maybe i am paranoid but i have doubts regarding how diffpatch lib is matching/comparing prices as order is not guaranteed from feed and API response and matching should always work.

butenkor:
Could you please also reply to my previous comments (https://github.com/sphereio/sphere-node-sdk/pull/251#discussion_r167601660, https://github.com/sphereio/sphere-node-sdk/pull/251#discussion_r167609359)

src/spec/sync/product-sync.spec.coffee
wizzy25: The first check is still the ID. This way, a scope can also be "updated" if the ID stays the same

src/spec/sync/product-sync.spec.coffee
wizzy25: Mixing the order will not trigger a `changePrice` action, but instead a `removePrice` and `addPrice`. This was supposed to be a quick fix and going ahead to check against all prices in the array regardless of their position will introduce too much logic to the code. That and the fact that this is soon-to-be deprecated makes me think it won't be worth the effort

src/coffee/sync/utils/product.coffee
wizzy25: In the case that a price scope is to be changed, it will be matched by ID

src/coffee/sync/utils/product.coffee
wizzy25: > Why price would be not an array and simultaneously have a array size greater than 0? _.isArray(price) and _.size(price)

Checking that it isn't an empty `object`, but yeah, I will add comments to the code

src/spec/sync/product-sync.spec.coffee
butenkor: Does new nodejs library support matching prices by scope already? If not maybe we should at least create an issue there?

wizzy25: The JS SDK doesn't support this. I will add a ticket for that

PR #255, opened 2018-03-06
build(node): drop support for node versions below 6

### LINKED ISSUE #254: Drop support for Node v0.x
### Description

Support for Node v0.x should be dropped, considering that LTS support has been dropped since 2016.
Dropping support for Node v4 should also be considered :-)

Here you can view the [LTS timetable](https://github.com/nodejs/Release#release-schedule)

What are your thoughts? @daern91 @junajan @butenkor

daern91:
resolves #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+# ⛔ [Deprecated]
+
+This repo has been deprecated and the code migrated to this [repo](https://github.com/commercetools/nodejs).
+
 ![commercetools logo](https://cdn.rawgit.com/commercetools/press-kit/master/PNG/72DPI/CT%20logo%20horizontal%20RGB%2072dpi.png)
 
 # Node.js SDK
@@ -9,8 +13,6 @@
 [commercetools](https://commercetools.com/) is a cloud-based commerce platform.
 
 Officially supported Node.js SDK library for working with the commercetools<span>&trade;</span> platform HTTP API, with OAuth2 support.
-
-> We are working on a [`v2.0`](https://github.com/sphereio/sphere-node-sdk/tree/rewrite-2.0) version, [more info](docs/MIGRATION-2.0.md) coming soon.
 
 ## Getting Started
 Install the module with `npm install sphere-node-sdk`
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "sphere-node-sdk",
   "description": "Officially supported Node.js SDK library for working with the SPHERE.IO HTTP API, with OAuth2 support.",
-  "version": "1.24.0",
+  "version": "1.25.0",
```
butenkor: Not major version?

daern91: Ah, sorry, I'm new to semver. Thought this would be breaking since anyone using old node versions need to update, @butenkor ? Would correct be `1.24.1` still?

butenkor: major would be 2.0.0

daern91: Got it. Thanks a lot, @butenkor ! :)

butenkor:
Maybe we should add deprecation notice into main README and link to new nodejs repository? I would like to avoid confusion with release of 2.0.0 as if this SDK is being actively developed. It should rather serve as clearly visible notice that it is a breaking change.

daern91:
Good idea. I've added that to readme now.

README.md
```diff
@@ -1,3 +1,7 @@
+# ⛔ [Deprecated]
+
+This repo has been deprecated and the code migrated to this [repo](https://github.com/commercetools/nodejs).
```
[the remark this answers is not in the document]
daern91: Sounds good @adnasa! Could you please check the link you sent above? It seems to direct to the nodejs repo and not to the `migration guide`. Then I'll add the correct one asap 👍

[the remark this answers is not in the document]
adnasa: https://commercetools.github.io/nodejs/sdk/upgrading-from-sphere-node-sdk.html

daern91:
For sure. Added another PR for that branch [here](https://github.com/sphereio/sphere-node-sdk/pull/257), @adnasa.

PR #268, opened 2018-11-20
fix(rest): set correct predicate in paged iteration

### LINKED ISSUE #267: Wrong where predicate inside paged recursion
### Description
The predicate created [here](https://github.com/sphereio/sphere-node-sdk/blob/276d47e8d8870894773c2d242efedbbc321fc7f5/src/coffee/connect/rest.coffee#L212) is wrong as it returns the following:
`masterVariant(sku in ("123456789")) or variants(sku in ("123456789")) and id > "id-one"`

Since `and` has precedence over `or` this one will always return true when the SKU is inside `masterVariant` and give us back all results, regardless of the `ID`. Thus, creating an infinite loop.

### Expected Behavior
The predicate should add parenthesis to ensure that 
`masterVariant(sku in ("123456789")) or variants(sku in ("123456789"))` 
gets executed first.

This way, the predicate will look like this:
`(masterVariant(sku in ("123456789")) or variants(sku in ("123456789"))) and id > "id-one"`
and only return the products with `id` above `"id-one"`

### Possible Solution
Change the string to include parenthesis, like this:
`wherePredicate = {where: if params.where then "(#{params.where})%20and%20#{lastIdPredicate}" else lastIdPredicate}`

daern91:
#### Summary
PR sets correct predicate string inside the paged function.

#### Description
More reading and details on this issue can be found here: https://github.com/sphereio/sphere-node-sdk/issues/267

resolves #267

### DIFF
--- a/src/coffee/connect/rest.coffee
+++ b/src/coffee/connect/rest.coffee
@@ -209,7 +209,7 @@ class Rest
       wherePredicate = if params.where then {where: params.where} else {}
       if lastId
         lastIdPredicate = encodeURIComponent("id > \"#{lastId}\"")
-        wherePredicate = {where: if params.where then "#{params.where}%20and%20#{lastIdPredicate}" else lastIdPredicate}
+        wherePredicate = {where: if params.where then "(#{params.where})%20and%20#{lastIdPredicate}" else lastIdPredicate}
       debug 'PAGED where predicate: %j', wherePredicate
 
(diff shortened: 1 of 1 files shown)


butenkor:
Don't we have same issue [here](https://github.com/commercetools/nodejs/blob/master/packages/sdk-client/src/client.js#L120) too?

PR #272, opened 2018-11-30
feat(product-sync): add support for changeMasterVariant, refactor product sync

### LINKED ISSUE #271: Rework product-sync and add support for changeMasterVariant action
### Description
The current `productSync` currently generates a change of `masterVariant` (eg from v1 to v2) with a set of these three update actions:
 - remove v2
 - override v1 with all information from v2 (taken from productDraft)
 - add v2 back as a normal variant (also taken from productDraft)

The problem in this approach is, that when it deletes v2, it also removes prices. Normally we would take them from productDraft but they can be missing there and in order to not lose any price info we have a blacklisted price action group. But because here we don't call `removePrice` action but `removeVariant` the blacklist does catch this and prices are removed.

Another problem can be, that when overriding a `masterVariant` with data from v2, we would filter out `removePrice` actions so v2 would have now prices from v1.

### Expected Behavior
It should just generate a `changeMasterVariant` action.

We should also match product variants by their id or sku or key instead of their order.

junajan:
#### Summary
Resolves #271 

#### Description
This PR changes how we generate product update actions. Before we used `jsondiffpatch` on the whole product. Now, when syncing variants, we match variants by `id/key/sku` and diff only matched variants which are present in both old and new product.

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation


src/spec/sync/product-sync.spec.coffee
```diff
@@ -109,13 +109,186 @@ NEW_PRODUCT =
...
+        masterVariant:
+          key: 'sku1'
+        variants: [
+          { key: 'sku3' }
+          { key: 'sku2' }
+        ]
+
+      expect(=> @sync.buildActions(newProduct, oldProduct)).toThrow new Error 'A variant must either have an ID or an SKU.'
+
+#    it 'should handle removing variant', ->
```
lojzatran: What about these tests?

src/coffee/sync/utils/product.coffee
```diff
@@ -95,7 +100,57 @@ class ProductUtils extends BaseUtils
       actions.push action if action
     actions
 
-  # Private: map categoryOrderHints actions
+  matchesByIdOrKeyOrSku: (variant1, variant2) ->
+    variant1 and variant2 and (
+      (!isNil(variant1.id) and variant1.id == variant2.id) or
+      (!isNil(variant1.key) and variant1.key == variant2.key) or
+      (!isNil(variant1.sku) and variant1.sku == variant2.sku)
```
daern91: this `isNil` threw me a little, is it not enough to do 
`variant1.id === variant2.id` 
or 
`variant1.id and variant1.id === variant2.id)`

?

src/coffee/sync/utils/product.coffee
```diff
@@ -95,7 +100,57 @@ class ProductUtils extends BaseUtils
...
+        else
+          throw new Error('ProductSync does need at least one of "id" or "sku" to generate a remove action')
+
+        actions.push(removeAction)
+    actions
+
+  buildAddVariantActions: (newVariant) ->
+    addAction = _.deepClone(newVariant)
+    delete addAction._NEW_ARRAY_INDEX
+    delete addAction._MATCH_CRITERIA
```
daern91: huh, where did these two come from? `_MATCH_CRITERIA` and `_NEW_ARRAY_INDEX`

src/coffee/sync/utils/product.coffee
```diff
@@ -126,48 +181,14 @@ class ProductUtils extends BaseUtils
...
-  # Private: map product references
+  # Map product references
   #
   # diff - {Object} The result of diff from `jsondiffpatch`
   # old_obj - {Object} The existing product
   # new_obj - {Object} The product to be updated
   #
   # Returns {Array} The list of actions, or empty if there are none
-  actionsMapReferences: (diff, old_obj, new_obj) ->
+  actionsMapReferences: (diff, new_obj, old_obj) ->
```
daern91: ahh, thanks for this one..  nice with some more consistency.

src/coffee/sync/utils/product.coffee
```diff
@@ -228,14 +249,14 @@ class ProductUtils extends BaseUtils
 
     _.sortBy actions, (a) -> a.action is 'addToCategory'
 
-  # Private: map product prices
+  # Map product prices
```
daern91: Im not aware of these ones. What does this change do?

src/spec/sync/product-sync.spec.coffee
```diff
@@ -136,20 +309,1026 @@ describe 'ProductSync', ->
...
+  #        { action: 'removeVariant', id: 5 }
+  #        { action: 'removeVariant', id: 6 }
+  #        { action: 'removeVariant', id: 7 }
+  #        { action: 'addVariant', sku: 'SKUadded' }
+  #        { action: 'addVariant', sku: 'SKUchanged!' }
+  #        { action: 'addVariant', attributes: [ { name: 'foo', value: 'bar' } ] }
+  #        { action: 'addVariant', attributes: [ { name: 'some', value: 'thing' } ] }
+  #        { action: 'addVariant', attributes: [ { name: 'yet', value: 'another' } ] }
+  #        { action: 'addVariant', sku: 'v10', attributes: [ { name: 'something', value: 'else' } ] }
+  #        { action: 'addVariant', sku: 'SKUwins' }
```
daern91: this `todo`?

src/coffee/sync/product-sync.coffee
```diff
@@ -36,23 +36,105 @@ class ProductSync extends BaseSync
...
+    # - add new variants
+    # - promote a new masterVariant
+    # - remove masterVariant
+
+    # divide removeVariant actions to two groups - one for masterVariant, second for normal variants
+    partitionedRemoveActionsByMasterVariant = _.partition variantActionGroups.removeVariantActions, (action) =>
+      oldObj.masterVariant and @_utils.matchesByIdOrKeyOrSku(action, oldObj.masterVariant)
+
+    allActions.push @_mapActionOrNot 'variants', => partitionedRemoveActionsByMasterVariant[1] # normal variants
+    allActions.push variantActionGroups.variantUpdateActions # already white/black listed from @_diffVariant
```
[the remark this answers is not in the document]
junajan: Based on the discussion from the last week, we should first update existing variants. Because if you have some `sameForAll` attribute, we should first update existing variants before we try to add new variants with this new value - otherwise we would run into an error because existing variants would still have the old value

src/coffee/sync/product-sync.coffee
```diff
@@ -36,23 +36,105 @@ class ProductSync extends BaseSync
     @sameForAllAttributeNames = sameForAllAttributeNames
     super new_obj, old_obj
 
-  _doMapActions: (diff, new_obj, old_obj) ->
+  _ensureDefaultProperties: (variant) ->
```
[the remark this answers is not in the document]
junajan: Hi @lojzatran  .. it looks like the property `actionGroups`  is used somewhere else because other sync classes have this property too.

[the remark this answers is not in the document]
junajan: What do you think is wrong on this class comment?
```
# Public: Define a `ProductSync` to provide basic methods to sync SPHERE.IO products.
#
# Action groups for products are:
# - `base` (name, slug, description)
# - `references` (taxCategory)
# - `prices`
# - `attributes`
# - `images`
# - `variants`
# - `categories`
#
# Examples
#
#   {ProductSync} = require 'sphere-node-sdk'
#   sync = new ProductSync
#   syncedActions = sync.buildActions(newProduct, existingProduct)
#   if syncedActions.shouldUpdate()
#     client.products.byId(syncedActions.getUpdatedId())
#     .update(syncedActions.getUpdatePayload())
#   else
#     # do nothing
```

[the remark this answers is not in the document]
lojzatran: OK then in this case this comment is good.

src/coffee/sync/utils/product.coffee
```diff
@@ -228,14 +249,14 @@ class ProductUtils extends BaseUtils
...
   #
-  # diff - {Object} The result of diff from `jsondiffpatch`
-  # old_obj - {Object} The existing product
-  # new_obj - {Object} The product to be updated
+  # prices - {Object} The result of variant diff from `jsondiffpatch`
```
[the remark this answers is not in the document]
junajan: I updated the variable name here 01fd315

src/coffee/sync/utils/product.coffee
```diff
@@ -272,96 +293,25 @@ class ProductUtils extends BaseUtils
...
-  # Private: map product attributes
+    actions
+
+  # Map product attributes
   #
-  # diff - {Object} The result of diff from `jsondiffpatch`
-  # old_obj - {Object} The existing product
-  # new_obj - {Object} The product to be updated
-  # sameForAllAttributeNames - {Array} A list of names of `SameForAll` attributes
+  # attributes - {Object} The result of variant diff from `jsondiffpatch`
```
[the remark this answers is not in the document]
junajan: updated here 01fd315

src/spec/sync/product-sync.spec.coffee
junajan: I removed them here 01fd315.. the functionality of adding and removing variants is already tested in other tests

src/coffee/sync/product-sync.coffee
```diff
@@ -36,23 +36,105 @@ class ProductSync extends BaseSync
...
+    # add default properties so we always compare arrays to arrays
+    @_ensureDefaultProperties(newVariant)
+    @_ensureDefaultProperties(oldVariant)
+
+    actions = []
+    diff = @_utils.diff(oldVariant, newVariant)
+
+    if diff
+      actions = []
+        # TODO check action group for base variant update actions - original was attributes
```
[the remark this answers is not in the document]
junajan: Removed todo here - 01fd315 .. these actions for updating sku/key were under `attributes` action group before so it is I guess ok if we keep it that way ..

src/coffee/sync/utils/product.coffee
junajan: If we don't use `isNil`, it would match incorrectly variants where existing and the new variant id is undefined even when skus and keys are different.

If we change it to `variant1.id and variant1.id === variant2.id` it won't work when id is `0`.

src/coffee/sync/utils/product.coffee
junajan: just removes private flag from documentation I guess.. because we call this method from outside

src/coffee/sync/utils/product.coffee
daern91: ahh.. I see. Thanks for explaining.

PR #273, opened 2018-12-06
fix(product-sync): fix matching variants

junajan:
#### Summary
Fix how we match variants.

#### Description
Originally, we were matching variants in a single loop by comparing id, sku or key.
Because we used id as a primary matching condition, there was happening, that we matched variants with the same ids but with different skus which later failed due to a duplicate slug error.

For example when syncing these two products:
```js
// existing product
{
	masterVariant: {
		id: 1,
		sku: "sku1"
	},
	variants: [{
		id: 2,
		sku: "sku3"
	}, {
		id: 3,
		sku: "sku2"
	}]
}
// and new product
{
	masterVariant: {
		id: 1,
		sku: "sku1"
	},
	variants: [{
		id: 2,
		sku: "sku2"
	}, {
		id: 3,
		sku: "sku3"
	}]
}
```
It would match variants using id and generate update actions:
```js
[{
	action: "setSku",
	variantId: 2,
	sku: "sku2"
}, {
	action: "setSku",
	variantId: 3,
	sku: "sku3"
}]
```
Which would fail because we can't set `sku2` to the first variant because it already exists. Therefore we should match by skus, then by keys and then, as a fallback, by id.

#### Todo

- Tests
    - [x] Unit
    - [ ] Integration
    - [ ] Acceptance
- [ ] Documentation

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -101,9 +101,7 @@ class ProductUtils extends BaseUtils
     actions
 
   matchesBySkuOrKeyOrId: (variant1, variant2) ->
-    @matchesBySku(variant1, variant2) or
-      @matchesByKey(variant1, variant2) or
-      @matchesById(variant1, variant2)
+    @matchesBySku(variant1, variant2) or @matchesByKey(variant1, variant2) or @matchesById(variant1, variant2)
 
   matchesById: (variant1, variant2) ->
     variant1 and variant2 and not isNil(variant1.id) and variant1.id == variant2.id
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -124,10 +124,10 @@ class ProductUtils extends BaseUtils
       action =
         action: 'changeMasterVariant'
 
-      if newMasterVariant.id
-        action.variantId = newMasterVariant.id
-      else if newMasterVariant.sku
+      if newMasterVariant.sku
         action.sku = newMasterVariant.sku
+      else if newMasterVariant.id
+        action.variantId = newMasterVariant.id
       else
         throw new Error(
           'ProductSync needs at least one of "id" or "sku" to generate changeMasterVariant update action'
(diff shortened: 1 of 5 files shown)


src/coffee/sync/utils/product.coffee
```diff
@@ -102,14 +102,26 @@ class ProductUtils extends BaseUtils
 
   matchesByIdOrKeyOrSku: (variant1, variant2) ->
     variant1 and variant2 and (
-      (!isNil(variant1.id) and variant1.id == variant2.id) or
+      (!isNil(variant1.sku) and variant1.sku == variant2.sku) or
```
lojzatran: Can you use `matchesBySku` function here?

junajan: please check here 93c1eb0

src/coffee/sync/utils/product.coffee
```diff
@@ -102,14 +102,26 @@ class ProductUtils extends BaseUtils
 
   matchesByIdOrKeyOrSku: (variant1, variant2) ->
```
lojzatran:
```suggestion
  matchesBySkuOrKeyOrId: (variant1, variant2) ->
```

junajan: 6f45d0e

junajan:
Waiting for coverage reports - there is reported a degraded performance on coveralls.io status page - http://status.coveralls.io/

src/coffee/sync/utils/product.coffee
```diff
@@ -100,19 +100,27 @@ class ProductUtils extends BaseUtils
...
-    variantList.find (testedVariant) =>
-      @matchesByIdOrKeyOrSku(testedVariant, variant)
+    return variantList.find((oldVariant) => @matchesBySku(variant, oldVariant)) or
+      variantList.find((oldVariant) => @matchesByKey(variant, oldVariant)) or
+      variantList.find((oldVariant) => @matchesById(variant, oldVariant)) or
+      undefined # if not found, return undefined
 
   buildChangeMasterVariantAction: (newMasterVariant, oldMasterVariant) ->
-    if newMasterVariant and oldMasterVariant and not @matchesByIdOrKeyOrSku(newMasterVariant, oldMasterVariant)
+    if newMasterVariant and oldMasterVariant and not @matchesBySkuOrKeyOrId(newMasterVariant, oldMasterVariant)
```
butenkor: Should not we change the order also here and [here](https://github.com/sphereio/sphere-node-sdk/pull/273/files#diff-a94247bc9925c68272e1e3c1213c3556L119) and [here](https://github.com/sphereio/sphere-node-sdk/pull/273/files#diff-a94247bc9925c68272e1e3c1213c3556R145)? I think it might be better to always use sku if defined and fallback on internal only when nothing else possible.

junajan: Hi @butenkor .. updated here c8afbe7

PR #302, opened 2019-07-26
feat: add method to get version number

sebbsan:
#### Summary
added method that returns the version of this SDK

#### Description
This was necessary to remove dependency in sphere-node-utils to an unmaintained library that introduces vulnerabilities

#### Todo

- Tests
    - [x ] Unit
    - [x ] Integration
    - [x ] Acceptance
- [ ] Documentation

### DIFF
--- a/src/coffee/utils.coffee
+++ b/src/coffee/utils.coffee
@@ -1,5 +1,4 @@
 _ = require 'underscore'
-appRoot = require('app-root-path')
 
 module.exports =
 
@@ -79,5 +78,5 @@ module.exports =
    * @return {Object} the version number as determined from package.json
   ###
   getVersion: ->
-    pjson = require appRoot + '/package.json'
+    pjson = require '../package.json'
     return pjson.version
--- a/src/coffee/utils.coffee
+++ b/src/coffee/utils.coffee
@@ -72,3 +72,10 @@ module.exports =
   _censorHeaderObj: (header) ->
     header.Authorization = "Bearer **********"
     return header
+
+  ###*
+   * Returns the version of this SDK 
+   * @return {Object} the version number as determined from package.json
+  ###
+  getVersion: () ->
+    return process.env.npm_package_version
--- a/src/coffee/utils.coffee
+++ b/src/coffee/utils.coffee
@@ -1,4 +1,5 @@
 _ = require 'underscore'
+appRoot = require('app-root-path')
 
 module.exports =
 
@@ -78,4 +79,5 @@ module.exports =
    * @return {Object} the version number as determined from package.json
   ###
   getVersion: () ->
-    return process.env.npm_package_version
+    pjson = require appRoot + '/package.json'
+    return pjson.version
--- a/src/coffee/utils.coffee
+++ b/src/coffee/utils.coffee
@@ -75,9 +75,9 @@ module.exports =
     return header
 
   ###*
-   * Returns the version of this SDK 
+   * Returns the version of this SDK
    * @return {Object} the version number as determined from package.json
   ###
-  getVersion: () ->
+  getVersion: ->
     pjson = require appRoot + '/package.json'
     return pjson.version
--- a/src/spec/utils.spec.coffee
+++ b/src/spec/utils.spec.coffee
@@ -1,6 +1,6 @@
 _ = require 'underscore'
 Utils = require '../lib/utils'
-appRoot = require('app-root-path')
+Semver = require 'semver'
 
 describe 'Utils', ->
 
@@ -57,6 +57,6 @@ describe 'Utils', ->
   describe ':: getVersion', ->
 
     it "should return the current version of the sphere node sdk", ->
-      pjson = require appRoot + '/package.json'
+      pjson = require '../package.json'
       expect(Utils.getVersion()).toBe pjson.version
-      expect(Utils.getVersion()).toMatch(/^(\d+\.)?(\d+\.)?(\*|\d+)$/)
+      expect(Semver.valid(Utils.getVersion())).toBeString
--- a/src/spec/utils.spec.coffee
+++ b/src/spec/utils.spec.coffee
@@ -52,3 +52,8 @@ describe 'Utils', ->
     ], (o) ->
       it "should get time in milliseconds for '#{o.type}'", ->
         expect(Utils.getTime(o.amount, o.type)).toBe o.expected_time
+
+  describe ':: getVersion', ->
+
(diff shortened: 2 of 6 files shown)


src/coffee/utils.coffee
```diff
@@ -1,4 +1,5 @@
 _ = require 'underscore'
+appRoot = require('app-root-path')
```
jenschude: Why adding a new dependency when you could just directly read the package.json?

E.G.: https://github.com/commercetools/nodejs/blob/59cb4a79f778efa4d4c4485c56311c1d0285ba65/packages/category-exporter/src/main.js

LEQADA:
@sebbsan do I understand correct that you want to remove some of the dependencies of `sphere-node-sdk` in the `sphere-node-utils` if version of the first one is lower than some N?

src/coffee/utils.coffee
sebbsan: this is traditionally a pain in the ass in NodeJS - there is no proper way to do it, see https://stackoverflow.com/questions/10265798/determine-project-root-from-a-running-node-js-application  - yes blame JavaScript ;)  Especially if you have to deal with cross-compilation like CoffeeScript - where the hell does my compiled JS end up? I have no clue at coding time ;) 

Adding this dependency is the lazy developer way to do it, plus this dependency doesn't introduce new ones, so I think it's ok to add it.

sebbsan:
@LEQADA the problem is this code here which I want to replace with this API method: 
https://github.com/sphereio/sphere-node-utils/blob/e39b07a188b081346b348bcf31ab95349e300ce4/src/coffee/helpers/user-agent.coffee#L6 

the dependency `load-module-pkg` introduces a vulnerability + hasn't been maintained for 2 years.
So once we release a new version of this, I can make a new version of `sphere-node-utils`

src/coffee/utils.coffee
daern91: I agree with Jens here. Looking at the spec file we can see how we import the built util file here:
https://github.com/sphereio/sphere-node-sdk/blob/c0113d74a7555644dd98e7ad7eaef284f0c847c1/src/spec/utils.spec.coffee#L2

which means we should be able to import it here as:
```
getVersion: () ->
  pjson = require '../package.json'
```

and inside the spec as
```
it "should return the current version of the sphere node sdk", ->
  pjson = require '../package.json'
```

daern91:
Thanks a lot for the PR. I first didn't understand why we needed this but after reading your response to @LEQADA I think it makes sense.

src/coffee/utils.coffee
sebbsan: hmm I don't like this.. the coding time file system does not match the runtime file system..

sebbsan: `util.coffee` is in `./src/coffee/...` - so at coding time I would have to assume that the `package.json` lives in `../../package.json` , after cross compilation the file ends up in `./lib` and therefore `../package.json` works, you see what I mean? So this code works incidentally, but it's not what I would call "systematic" because it assumes something about the file system which is configured in a build script. If I would change the build script, this library would break all over the place.

daern91: I wouldn't call it incidental. It's how the mapping is set up in the Grunt/config file here:
https://github.com/sphereio/sphere-node-utils/blob/b3c15aea8fb8fd85e2356b87082d6c62bb53465e/Gruntfile.coffee#L28-L35

Meaning, you're right that we need to check how we've mapped the compiler in the project before starting to import things outside of the `src` folder. But that is a common procedure in node projects nowadays IMO. The same thing happens when using TypeScript/Flow/Babel.

It does, also, follow the pattern used in other files requiring modules from the outside.

E.g.
https://github.com/sphereio/sphere-node-utils/blob/4b691775646724af748d4cdd1e21d3467bec0253/src/spec/exports.spec.coffee#L1
https://github.com/sphereio/sphere-node-utils/blob/3b38c6393d55585df5b0a5d5b8a36f6be574b591/src/spec/helpers/project-credentials-config.spec.coffee#L3
https://github.com/sphereio/sphere-node-utils/blob/b3c15aea8fb8fd85e2356b87082d6c62bb53465e/src/spec/helpers/sftp.spec.coffee#L5-L6

sebbsan: Well, in my humble opinion, it's a strong coupling between code and the way the lib is packaged, and therefore is detrimental to coupling. 
But fine I will do it, under protest ;)

daern91: Ideally, I think it would be best if Grunt also supported some sort of module resolution. That way we simply wouldn't have to care about this at all and could always use relative paths from the coffee code.

Anyway, this is such old legacy repos that I bet no one would like to dive in to try to solve it 😄

src/spec/utils.spec.coffee
```diff
@@ -52,3 +53,10 @@ describe 'Utils', ->
     ], (o) ->
       it "should get time in milliseconds for '#{o.type}'", ->
         expect(Utils.getTime(o.amount, o.type)).toBe o.expected_time
+
+  describe ':: getVersion', ->
+
+    it "should return the current version of the sphere node sdk", ->
+      pjson = require appRoot + '/package.json'
+      expect(Utils.getVersion()).toBe pjson.version
+      expect(Utils.getVersion()).toMatch(/^(\d+\.)?(\d+\.)?(\*|\d+)$/)
```
LEQADA: According to npm semver `0.0.0-1` is a valid version. But this regex will fail here. Maybe it's better to add [semver](https://docs.npmjs.com/misc/semver) as a dependency and use it as a validator?

src/coffee/utils.coffee
butenkor: Be careful. We still use it in many production projects and it works ;)

sebbsan: NPM has built in resolution of package.json properties by adding everything as a environment variable...   but again this is a runtime dependency...

sebbsan: in the "good old times", we added stuff like version number during build time in properties files by the way

src/spec/utils.spec.coffee
sebbsan: ok I amended this - semver is now a dev dependency

src/coffee/utils.coffee
sebbsan: ok it's done in the new version

sebbsan:
@LEQADA @daern91 and after we have a new version of `sphere-node-utils` we need a new version of this one again. There is a circular dependency..

PR #304, opened 2019-08-16
fix(getVersion): add Utils to public interface of this module

sebbsan:
#### Summary
need Utils in the public interface to use it from sphere-node-utils

#### Description
Utils is exported, now it can be used externally. Because the new getVersion method is thought to be used externally. This is required to fix a vulnerability in sphere-node-utils that is spreading throughout the tools code base

#### Todo

- Tests
    - [x ] Unit
    - [ x] Integration
    - [x ] Acceptance
- [ ] Documentation


sebbsan:
sorry here is something I forgot in my last PR.

PR #324, opened 2020-03-06
feat(product-sync): on product variant actions, use SKU as the first option instead of variantId.

ahmetoz:
### Summary:
Pr is changing the product variant actions, now it's using SKU as the first option for variant update actions:

for example `setAttribute` action documentation explains it :
`variantId - Number or sku - String - Required`
 
#### Why we needed it?

We have lots of these kinds of errors: A staged variant with variantId '3' does not exist... I checked one of the product and it indeed doesn’t have that variant. I also checked the history of that product in master because I thought that the product could have before 3 variants and then one got removed. But it always has 2 variants. So it’s weird that we try to update with variant 3 when this variant never existed.

**It might not resolve the issue for us but could help with finding the reason.**

#### Error message from the API:
```
{
      action: {
       action: "setAttribute"        
       name: "attributeName"        
       staged: true        
       value: [
        0: "AU:0"         
        1: "DE:0"         
        2: "FR:0"         
        3: "NL:0"         
        4: "NZ:0"         
       ]
       variantId: 3        
      }
      actionIndex: 7       
      code: "InvalidOperation"       
      message: "A staged variant with variantId '3' does not exist."       
}
```

**Review Note**: I did not add additional tests for the case, currently the cases already covered, so I just updated the existing ones.

### DIFF
--- a/src/coffee/sync/utils/product.coffee
+++ b/src/coffee/sync/utils/product.coffee
@@ -124,14 +124,7 @@ class ProductUtils extends BaseUtils
       action =
         action: 'changeMasterVariant'
 
-      if newMasterVariant.sku
-        action.sku = newMasterVariant.sku
-      else if newMasterVariant.id
-        action.variantId = newMasterVariant.id
-      else
-        throw new Error(
-          'ProductSync needs at least one of "id" or "sku" to generate changeMasterVariant update action'
-        )
+      @_setSkuOrVariantIdToUpdateAction(newMasterVariant, action)
       action
 
   buildRemoveVariantActions: (newVariants, oldVariants) ->
@@ -402,8 +395,8 @@ class ProductUtils extends BaseUtils
       delete price._MATCH_CRITERIA
       action =
         action: 'addPrice'
-        variantId: old_variant.id
         price: price
+      @_setSkuOrVariantIdToUpdateAction(old_variant, action)
     action
 
   buildVariantImagesAction: (images, old_variant, new_variant) ->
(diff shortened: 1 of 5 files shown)


ahmetoz:
@VineetKumarKushwaha 🙏 thanks for the review. could we release it?

src/coffee/sync/utils/product.coffee
```diff
@@ -452,16 +444,16 @@ class ProductUtils extends BaseUtils
       delete image._MATCH_CRITERIA
       action =
         action: 'addExternalImage'
-        variantId: variant.id
         image: image
+      @_setSkuOrVariantIdToUpdateAction(variant, action)
```
butenkor: Can we use the same pattern for naming everywhere? I would suggest using either `old_variant` or `new_variant` ?

src/coffee/sync/utils/product.coffee
```diff
@@ -432,13 +425,12 @@ class ProductUtils extends BaseUtils
 
           else if _.has(image, 'label') and
           (image.label.length == 1 or image.label.length == 2)
-
-            actions.push(
+            action =
               action: 'changeImageLabel'
```
butenkor: I wonder if it is some outdated action not supported by API anymore as I can see only `setImageLabel` in the doc.

src/coffee/sync/utils/product.coffee
ahmetoz: https://github.com/sphereio/sphere-node-sdk/pull/324/commits/48976579d71080c3d783e187e5a8db2b74ea3ffe

src/coffee/sync/utils/product.coffee
ahmetoz: https://github.com/sphereio/sphere-node-sdk/pull/324/commits/e8462920f582e80ee06a138b90d91c59762a80f7

